PR #141, opened 2026-04-28
fix: checkpoint issuance backfills before serving

findolor:
## Motivation

Closes RAI-233.

On restart, the issuer can spend a long time backfilling historical chain ranges before it becomes usable, leaving HTTP endpoints and health checks unavailable. The worst case is not just slow backfill: the service can redo already-processed ranges because progress is derived from stale startup state instead of durable per-range progress.

This follows the checkpoint backfill pattern from `st0x.liquidity` PR #592: persist progress, resume from `checkpoint + 1`, avoid unnecessary reprocessing, and keep avoidable catch-up work out of the critical startup path.

## Solution

- Added a `redemption_backfill_checkpoints` table keyed by vault.
- Updated redemption transfer backfill to:
  - resume from `last_processed_block + 1`
  - use `backfill_start_block` only as the first-run/configured floor
  - persist progress only after the requested range succeeds
  - write checkpoints monotonically with `MAX(...)`
  - log vault/range/checkpoint/detected/skipped counts
- Moved redemption transfer backfill into a background task after startup recovery, so HTTP serving is not blocked by avoidable historical transfer catch-up.
- Updated receipt backfill to start from the block after `ReceiptInventory::last_backfilled_block`, avoiding checkpoint-block reprocessing.
- Added periodic receipt backfill during uptime so receipt checkpoints stay current across long-running processes.
- Added guarded live receipt checkpoint advancement:
  - successful live logs advance the checkpoint
  - stale checkpoint commands are ignored
  - failed or unpositioned live logs block later live checkpoint advancement so restart/periodic backfill cannot skip them
- Updated `SPEC.md` and `README.md` to document checkpointed receipt/redemption backfill behavior.

## Checks

By submitting this for review, I am confirming I have done the following:

Validation run:

- `nix develop -c cargo test -q receipt_inventory`
- `nix develop -c cargo test -q redemption::backfill`
- `nix develop -c cargo test -q --lib`
- `nix develop -c cargo clippy --workspace --all-targets --all-features -- -D clippy::all -D warnings`
- `nix develop -c cargo check --workspace`
- `git --no-pager diff --check`

## Summary by CodeRabbit

* **New Features**
  * Durable per-vault checkpoints for receipt and redemption backfills
  * Periodic receipt backfill alongside live monitoring (live events won't advance durable checkpoints)
  * Transfer backfill runs asynchronously in background at startup and retries on failure

* **Behavior Change**
  * Restarts resume from the last successfully processed block; periodic backfills advance durable checkpoints only after ordered success

* **Documentation**
  * Updated backfill, checkpointing, and startup reconciliation descriptions

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,7 +10,8 @@ use sqlite_es::{
     SqliteCqrs, SqliteEventRepository, SqliteViewRepository, sqlite_cqrs,
 };
 use sqlx::{Pool, Sqlite, sqlite::SqlitePoolOptions};
-use std::sync::Arc;
+use std::{sync::Arc, time::Duration};
+use tokio::time::MissedTickBehavior;
 use tracing::{debug, error, info, warn};
 use url::Url;
 
@@ -121,6 +122,9 @@ impl std::fmt::Display for Quantity {
 /// Maximum decimal places supported by Alpaca's tokenization API.
 /// Quantities with more decimals must be truncated before sending to Alpaca.
 pub(crate) const ALPACA_MAX_DECIMALS: u32 = 9;
+
+const RECEIPT_BACKFILL_REFRESH_INTERVAL: Duration = Duration::from_secs(60);
+const TRANSFER_BACKFILL_RETRY_INTERVAL: Duration = Duration::from_secs(60);
 
 impl Quantity {
     pub(crate) const fn new(value: Decimal) -> Self {
@@ -321,9 +325,6 @@ pub async fn initialize_rocket(
         );
     }
 
-    // Transfer backfill must run after receipt backfill so that receipt inventory
-    // is available for burn planning, and before live monitoring so that missed
-    // transfers during downtime are detected before new ones come in.
     let transfer_backfiller = TransferBackfiller {
         provider: blockchain_setup.provider.clone(),
         bot_wallet,
@@ -335,15 +336,8 @@ pub async fn initialize_rocket(
         burn_manager: managers.burn.clone(),
     };
 
-    run_all_transfer_backfills(
-        &vault_configs,
-        &transfer_backfiller,
-        config.backfill_start_block,
-    )
-    .await?;
-
     spawn_all_receipt_monitors(
-        blockchain_setup.provider,
+        blockchain_setup.provider.clone(),
         &vault_configs,
         &receipt_inventory.cqrs,
         bot_wallet,
@@ -361,6 +355,21 @@ pub async fn initialize_rocket(
         &managers.burn,
     )
     .await;
+
+    spawn_periodic_receipt_backfills(
+        blockchain_setup.provider.clone(),
+        vault_configs.clone(),
+        receipt_inventory.cqrs.clone(),
+        receipt_inventory.event_store.clone(),
+        bot_wallet,
+        config.backfill_start_block,
+    );
+
+    spawn_all_transfer_backfills(
+        vault_configs.clone(),
+        transfer_backfiller,
+        config.backfill_start_block,
+    );
 
     spawn_all_redemption_detectors(
         &config.rpc_url,
@@ -710,6 +719,7 @@ async fn run_redemption_recovery(
 }
 
 /// Configuration for a single vault, extracted from TokenizedAssetView.
+#[derive(Clone, Copy)]
 struct VaultBackfillConfig {
     vault: Address,
     receipt_contract: Address,
@@ -782,10 +792,10 @@ async fn run_single_vault_backfill<P: Provider + Clone>(
         .load_aggregate(&vault.to_string())
         .await?;
 
-    let from_block = aggregate_context
-        .aggregate()
-        .last_backfilled_block()
-        .unwrap_or(backfill_start_block);
+    let from_block = next_receipt_backfill_block(
+        aggregate_context.aggregate().last_backfilled_block(),
+        backfill_start_block,
+    )?;
 
     info!(
         underlying,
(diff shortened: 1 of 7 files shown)


findolor:
* **#142** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/142?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#141** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/141?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/141?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=141).

<h2></h2>
<sub>

An organization admin has enabled the [Graphite Merge Queue](https://graphite.dev/docs/graphite-merge-queue) in this repository.

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios (commented):
disregard this commment

src/redemption/backfill.rs
```diff
@@ -187,6 +206,73 @@ where
...
+    pool: &Pool<Sqlite>,
+    vault: Address,
+    configured_start_block: u64,
+) -> Result<u64, TransferBackfillError> {
+    load_backfill_checkpoint(pool, vault).await?.map_or(
+        Ok(configured_start_block),
+        |last_processed_block| {
+            let next_unprocessed_block = last_processed_block
+                .checked_add(1)
+                .map_or(last_processed_block, |block| block);
```
JuaniRios: **(MEDIUM) — Silent overflow violates Financial Data Integrity rules**

When `last_processed_block == u64::MAX`, `checked_add(1)` returns `None` and `map_or(last_processed_block, |block| block)` silently falls back to `u64::MAX` — re-scanning the same block forever.

Not achievable probably, but better to safeguard against it

Suggested fix: return an error on overflow, matching the receipt backfill pattern.

src/lib.rs
```diff
@@ -854,11 +963,38 @@ async fn run_all_transfer_backfills<P: Provider + Clone>(
...
+{
+    tokio::spawn(async move {
+        if let Err(error) = run_all_transfer_backfills(
+            &vault_configs,
+            &backfiller,
+            backfill_start_block,
+        )
+        .await
+        {
+            warn!(
```
JuaniRios: **(MEDIUM) — No retry for background transfer backfill**

If this fails partway through (e.g., after 3 of 5 vaults), the remaining vaults won't be retried until service restart. Receipt backfill gets periodic retry via `spawn_periodic_receipt_backfills`, but transfer backfill is fire-and-forget.

This shifts the "no missed redemptions" guarantee from deterministic (startup fails) to eventually-consistent (next restart). Consider adding periodic retry (mirroring receipt backfill) or at minimum escalating to `error!` since this represents potential loss of redemption detection.

src/redemption/backfill.rs
```diff
@@ -73,17 +77,26 @@ where
...
 
         if from_block > current_block {
             info!(
+                %vault,
                 from_block,
                 current_block,
                 "Transfer backfill skipped: from_block is ahead of current block"
             );
 
+            save_backfill_checkpoint(&self.pool, vault, current_block).await?;
```
JuaniRios: **(LOW) — Unnecessary checkpoint save in skip path**

When `from_block > current_block`, this save is a no-op due to the SQL `MAX()` guard, but `updated_at` is unconditionally updated — giving the false impression the checkpoint was just modified. Consider removing this call since no work was done.

src/receipt_inventory/monitor.rs
```diff
@@ -256,6 +260,108 @@ where
...
+        &self,
+        log: &'a Log,
+        event_name: &'static str,
+        blocked_checkpoint_block: &mut Option<u64>,
+        process: impl FnOnce(&'a Log) -> FutureType,
+    ) where
+        FutureType:
+            std::future::Future<Output = Result<(), ReceiptMonitorError>>,
+    {
+        let Some(block_number) = log.block_number else {
```
JuaniRios, line 268: **(LOW) — Silently drops events when `block_number` is None**

The early return skips the `process` callback entirely. In the old code, some handlers (e.g., `process_withdraw`) would still execute without needing `block_number`. This is safe because periodic backfill covers the gap, but worth a comment explaining why dropping the event is acceptable.

src/receipt_inventory/monitor.rs
```diff
@@ -256,6 +260,108 @@ where
...
+
+        match process(log).await {
+            Ok(()) => {
+                self.clear_blocked_checkpoint_if_caught_up(
+                    blocked_checkpoint_block,
+                )
+                .await;
+
+                if blocked_checkpoint_block
+                    .is_none_or(|blocked_block| block_number < blocked_block)
```
JuaniRios: ** (LOW) — Out-of-order WebSocket events could advance checkpoint past unprocessed blocks**

Events from different subscription streams don't arrive in strict block order (`tokio::select!` picks whichever is ready). If block 102 succeeds and advances the checkpoint, then block 100 fails, `blocked_checkpoint_block` is set to 100 but the durable checkpoint is already at 102.

Mitigated by periodic backfill (60s) and the aggregate's monotonic guard, but the interaction is non-obvious. A brief comment explaining the safety invariant would help.

src/receipt_inventory/mod.rs
```diff
@@ -696,6 +705,26 @@ mod tests {
         ));
     }
 
+    #[tokio::test]
+    async fn test_advance_backfill_checkpoint_ignores_stale_block() {
```
JuaniRios: **#8 (LOW) — Missing boundary test for equal block number**

This tests `block_number: 99` against `last_backfilled_block: Some(100)` (strictly less). The guard uses `<=`, which also rejects equal values. Consider adding an assertion for `block_number: 100` — that's the common scenario (multiple logs in the same block each trying to advance).

JuaniRios (changes requested):
The PR achieves its goals well — durable checkpointing, background transfer backfill, periodic receipt backfill, and guarded live checkpoint advancement. Architecture is sound and test coverage is thorough.

There's some easy fixes we can do that i'd like to see before merging, this includes the findings below, and the ones attached to specific diffs.

Findings not anchored to a specific diff line:

** (LOW) — Missing happy-path tests for next_receipt_backfill_block** (src/lib.rs): Only overflow is tested. The redemption version has (None, 50)->50, (Some(80), 50)->81, (Some(20), 50)->50 — receipt should match.

(NIT) — Missing #[traced_test] on new backfill integration tests (src/redemption/backfill.rs): Per AGENTS.md, tests exercising business logic must assert on log output. backfill_persists_checkpoint_after_success and backfill_restart_skips_already_checkpointed_range exercise the full flow with INFO-level logs but have no tracing assertions.

(NIT) — Stale doc comment on run_all_transfer_backfills (src/lib.rs:931-933): Says "Runs after recovery... before Rocket is returned only as a spawned background task" — describes invocation context that's now the caller's concern, not this function's.

JuaniRios (approved):

hardyjosh (approved):

PR #181, opened 2026-06-15
feat: asset freeze status endpoint

0xgleb:
## Motivation

The liquidity rebalance guard (RAI-1038) needs to skip frozen assets before starting a rebalancing flow, so it needs an issuance-side endpoint to read a per-asset freeze status. Part of RAI-1034 (issuance bot asset freezing); implements RAI-1037.

## Solution

Adds `GET /tokenized-assets/<underlying>/status` (internal auth) returning `{ underlying, enabled, frozen }`, read from `tokenized_asset_view` via the typed view helper. A dedicated endpoint keeps the liquidity guard's polling contract decoupled from the fuller asset-detail payload, which can evolve independently. `enabled` stays true for a frozen asset (frozen != de-listed); `frozen` reflects the freeze state. Documented in SPEC.md.

Stacked on #180 (RAI-1035), which introduces the freeze state this endpoint reads.

## Checks

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

PR #180, opened 2026-06-15
feat: frozen asset state

0xgleb:
## Motivation

Around a dividend ex-date the issuer must stop creating new supply for an asset (RAI-1034). This adds the CQRS freeze/unfreeze state machine and mint-side enforcement — the foundation for the single dividend freeze/unfreeze mechanism (issuance freeze + the liquidity rebalance guard). Freeze is modeled as its own state, orthogonal to listing, so in-flight redemptions still complete. Implements RAI-1035.

## Solution

- `TokenizedAsset` gains an `AssetStatus` (`Enabled` | `Frozen`) replacing the always-true `enabled` flag; new idempotent `Freeze` / `Unfreeze` commands and `Frozen` / `Unfrozen` events.
- `POST /inkind/issuance` rejects a frozen asset with a new typed `AssetFrozen` error, distinct from `AssetNotAvailable`, so the suspension is observable and not conflated with de-listing.
- A frozen asset stays in `list_enabled_assets()` and keeps resolving its vault, so redemption detection and burns of a frozen asset keep working (proven by test).
- View migration drops + recreates `tokenized_asset_view` so catch_up rebuilds rows with the new `status` field. SPEC.md state machine updated.
- Also tightens AGENTS.md's sqlx-DB-error guidance (use `sqlx db reset -y`, don't improvise).

## Checks

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

## Summary by CodeRabbit

* **New Features**
  * Assets can now be frozen to prevent minting while maintaining redemption functionality

* **Bug Fixes**
  * Mint endpoint now properly rejects minting requests for frozen assets with a clear error response

* **Refactor**
  * Updated internal asset status model to support frozen state alongside enabled state

PR #182, opened 2026-06-16
feat: asset freezing

0xgleb:
## Motivation

The dividend freeze needs an M1 manual trigger on the issuer host to stop new mints for an asset around an ex-date (RAI-1034). Implements RAI-1036.

## Solution

Adds a dedicated `issuer` binary (`src/bin/issuer.rs`), separate from the HTTP server binary — `issuer freeze <UNDERLYING>` / `issuer unfreeze <UNDERLYING>` / `issuer status <UNDERLYING>`. `src/main.rs` stays the pure server. The CLI opens the same SQLite store and dispatches the CQRS `Freeze` / `Unfreeze` commands through the event-sorcery `Store` (never writing the `events` table directly); each mutating subcommand prints the resolved asset's current status, requires confirmation, and reports the idempotent outcome. Dispatch logic lives in a testable lib module (`AssetAdmin` + `run_asset_command`); the binary owns clap parsing + the stdin confirmation. Future issuer actions (`mint`, `donate`) become additional subcommands here. Documented in SPEC.md.

Stacked on #181 (RAI-1037) and #180 (RAI-1035).

## Checks

## Summary by CodeRabbit

* **New Features**
  * Issuer CLI tool for manually triggering tokenized asset freeze and unfreeze operations on the issuer host
  * Asset freeze status query functionality
  * Interactive confirmation prompts before executing state-changing commands
  * Idempotent operations with clear user-friendly feedback for state transitions

### DIFF
--- a/src/bin/issuer.rs
+++ b/src/bin/issuer.rs
@@ -1,0 +1,8 @@
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    // `dotenv` (not `dotenv_override`) so a stale `.env` on the issuer host can't
+    // override host-provided env (e.g. DATABASE_URL) and point the CLI at the
+    // wrong database.
+    dotenvy::dotenv().ok();
+    st0x_issuance::run_issuer_cli().await
+}
--- a/src/tokenized_asset/cli.rs
+++ b/src/tokenized_asset/cli.rs
@@ -1,0 +1,652 @@
+use clap::{Args, Parser, Subcommand};
+use event_sorcery::{
+    AggregateError, LifecycleError, ReconcileError, Store, StoreBuilder,
+};
+use sqlx::sqlite::{SqliteConnectOptions, SqlitePoolOptions};
+use sqlx::{Pool, Sqlite};
+use std::io::{self, Write};
+use std::str::FromStr;
+use std::sync::Arc;
+use std::time::Duration;
+
+use super::view::{TokenizedAssetViewError, load_asset_by_underlying};
+use super::{
+    AssetStatus, TokenizedAsset, TokenizedAssetCommand, UnderlyingSymbol,
+};
+use crate::config::{
+    DEFAULT_DATABASE_MAX_CONNECTIONS, DEFAULT_DATABASE_URL, LogLevel,
+    setup_tracing,
+};
+
+/// Parses and runs the issuer-host CLI end to end. The `issuer` binary is a thin
+/// wrapper over this entry point.
+///
+/// # Errors
+///
+/// Returns an error if argument parsing fails, the store cannot be opened, the
+/// asset is not supported, the operator aborts a mutation, or the command
+/// dispatch fails.
+pub async fn run_issuer_cli() -> anyhow::Result<()> {
+    setup_tracing(&LogLevel::Info);
+    IssuerCli::parse().dispatch().await
+}
+
+#[derive(Parser)]
+#[command(
+    name = "issuer",
+    version,
+    about = "Issuer-host admin CLI for st0x.issuance"
+)]
+struct IssuerCli {
+    #[command(subcommand)]
+    command: IssuerCommand,
+}
+
+#[derive(Subcommand)]
+enum IssuerCommand {
+    /// Freeze an asset: reject new mints (in-flight redemptions still complete).
+    Freeze(AssetArgs),
+    /// Unfreeze an asset: resume accepting new mints.
+    Unfreeze(AssetArgs),
+    /// Print an asset's current freeze status.
+    Status(AssetArgs),
+}
+
+#[derive(Args)]
+struct AssetArgs {
+    /// Underlying symbol, e.g. SGOV.
+    #[arg(value_parser = parse_underlying)]
+    underlying: UnderlyingSymbol,
+    #[arg(
+        long = "database-url",
+        env = "DATABASE_URL",
+        default_value = DEFAULT_DATABASE_URL,
+        value_parser = parse_sqlite_url
+    )]
+    database_url: String,
+    #[arg(
+        long,
+        env = "DATABASE_MAX_CONNECTIONS",
+        default_value_t = DEFAULT_DATABASE_MAX_CONNECTIONS,
+        value_parser = clap::value_parser!(u32).range(1..)
+    )]
+    database_max_connections: u32,
+}
+
+impl IssuerCli {
+    async fn dispatch(self) -> anyhow::Result<()> {
+        match self.command {
+            IssuerCommand::Freeze(args) => {
+                run_asset_command(AssetAction::Freeze, &args).await
+            }
+            IssuerCommand::Unfreeze(args) => {
+                run_asset_command(AssetAction::Unfreeze, &args).await
+            }
+            IssuerCommand::Status(args) => {
+                run_asset_command(AssetAction::Status, &args).await
+            }
+        }
+    }
+}
+
+enum AssetAction {
+    Freeze,
+    Unfreeze,
+    Status,
+}
+
+/// Connects to the store, prints the resolved database so the operator can
+/// confirm they are acting on the intended store, and runs the action with the
+/// real stdin confirmation prompt.
+async fn run_asset_command(
+    action: AssetAction,
+    args: &AssetArgs,
+) -> anyhow::Result<()> {
+    println!("Using database: {}", args.database_url);
+
+    let admin =
+        AssetAdmin::connect(&args.database_url, args.database_max_connections)
+            .await?;
+
+    execute(&admin, action, &args.underlying, prompt_confirm).await
+}
+
+/// Orchestrates a single action against an already-connected admin. The
+/// confirmation is injected so the abort/confirm branches are unit-testable
+/// without driving real stdin. Aborting a mutation returns an error (non-zero
+/// exit) so automation can distinguish "operator declined" from "done".
+async fn execute(
+    admin: &AssetAdmin,
+    action: AssetAction,
+    underlying: &UnderlyingSymbol,
+    confirm: impl Fn(&str) -> io::Result<bool>,
+) -> anyhow::Result<()> {
+    // Display the current status for the operator to confirm against, and reject
+    // an unknown asset up front. This snapshot only drives the prompt and the
+    // not-found check — the freeze/unfreeze decision is NOT derived from it (see
+    // `freeze`/`unfreeze`), so a concurrent write landing in the confirmation
+    // window can never leave the asset in the wrong persisted state.
+    let report = admin.status(underlying).await?.ok_or_else(|| {
+        AssetAdminError::NotFound { underlying: underlying.clone() }
+    })?;
+    println!("{report}");
+
+    match action {
+        AssetAction::Status => Ok(()),
+        AssetAction::Freeze => {
+            if !confirm(&format!("Freeze {underlying}?"))? {
+                anyhow::bail!("aborted by operator");
+            }
+            match admin.freeze(underlying).await? {
+                FreezeOutcome::Froze => println!("Froze {underlying}."),
+                FreezeOutcome::AlreadyFrozen => {
+                    println!("{underlying} was already frozen.");
+                }
+            }
+            Ok(())
+        }
+        AssetAction::Unfreeze => {
+            if !confirm(&format!("Unfreeze {underlying}?"))? {
+                anyhow::bail!("aborted by operator");
+            }
+            match admin.unfreeze(underlying).await? {
+                UnfreezeOutcome::Unfroze => println!("Unfroze {underlying}."),
+                UnfreezeOutcome::AlreadyEnabled => {
+                    println!("{underlying} was already enabled.");
+                }
+            }
+            Ok(())
+        }
+    }
+}
+
+/// Issuer-host admin for freezing/unfreezing supported assets.
+///
+/// Opens the same SQLite event store the server uses and dispatches the CQRS
+/// `Freeze` / `Unfreeze` commands through the event-sorcery `Store` — never
+/// writing the `events` table directly.
+pub(crate) struct AssetAdmin {
+    store: Arc<Store<TokenizedAsset>>,
+    pool: Pool<Sqlite>,
+}
+
+/// Outcome of a freeze request, so the caller can report an idempotent no-op
+/// distinctly from an actual state change. A missing asset is an
+/// `AssetAdminError::NotFound`, not an outcome: `execute` rejects unknown assets
+/// up front, so `freeze` only runs against an asset that exists.
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum FreezeOutcome {
+    Froze,
+    AlreadyFrozen,
+}
+
+/// Outcome of an unfreeze request. A missing asset is an
+/// `AssetAdminError::NotFound`, not an outcome (see `FreezeOutcome`).
+#[derive(Debug, PartialEq, Eq)]
+pub(crate) enum UnfreezeOutcome {
+    Unfroze,
+    AlreadyEnabled,
+}
+
+/// A supported asset's freeze status, formatted for the CLI.
+#[derive(Debug)]
+pub(crate) struct AssetStatusReport {
+    pub(crate) underlying: UnderlyingSymbol,
+    pub(crate) status: AssetStatus,
+}
+
+impl std::fmt::Display for AssetStatusReport {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let state = match self.status {
+            AssetStatus::Frozen => "frozen",
+            AssetStatus::Enabled => "enabled",
+        };
+        write!(f, "{} is {state}", self.underlying)
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum AssetAdminError {
+    #[error("database error: {0}")]
+    Database(#[from] sqlx::Error),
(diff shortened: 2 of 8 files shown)


0xgleb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/182?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#189** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/189?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#188** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/188?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#186** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/186?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#185** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/185?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#184** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/184?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#183** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/183?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#182** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/182?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/182?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#181** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/181?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#196** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=182).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

src/bin/issuer.rs
```diff
@@ -0,0 +1,8 @@
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    // `dotenv` (not `dotenv_override`) so a stale `.env` on the issuer host can't
+    // override host-provided env (e.g. DATABASE_URL) and point the CLI at the
+    // wrong database.
+    dotenvy::dotenv().ok();
+    st0x_issuance::run_issuer_cli().await
```
JuaniRios, line 7: should fix: this new binary never gets copied into the production image. The Dockerfile only patchelfs and copies target/${BUILD_PROFILE}/st0x-issuance as ./server, so the workflow documented in SPEC.md (docker compose exec issuance-bot ./issuer freeze SGOV) will fail because ./issuer isn't there. Add a patchelf plus COPY for the issuer binary into the final image.

src/bin/issuer.rs
```diff
@@ -0,0 +1,8 @@
+#[tokio::main]
```
JuaniRios, line 1: should fix: adding a second binary target makes plain cargo run ambiguous because there's no default-run set, which breaks the cargo run command documented in AGENTS.md and README. Add default-run = "st0x-issuance" under [package] in Cargo.toml.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+        default_value = DEFAULT_DATABASE_URL,
+        value_parser = parse_sqlite_url
+    )]
+    db: String,
+    #[arg(
+        long,
+        env = "DATABASE_MAX_CONNECTIONS",
+        default_value_t = DEFAULT_DATABASE_MAX_CONNECTIONS
+    )]
+    database_max_connections: u32,
```
JuaniRios, line 73: should fix: this is an unconstrained u32 that flows straight into SqlitePoolOptions::max_connections. A value of 0 reaches crossbeam's ArrayQueue::new(0), which asserts capacity > 0 and panics, so a bad env var or arg aborts the command instead of giving a clean error. Constrain it with value_parser = clap::value_parser!(u32).range(1..), or parse as NonZeroU32.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+        &self,
+        underlying: &UnderlyingSymbol,
+    ) -> Result<FreezeOutcome, AssetAdminError> {
+        match self.current_status(underlying).await? {
+            None => Err(AssetAdminError::NotFound {
+                underlying: underlying.clone(),
+            }),
+            Some(AssetStatus::Frozen) => Ok(FreezeOutcome::AlreadyFrozen),
+            Some(AssetStatus::Enabled) => {
+                self.dispatch(underlying, TokenizedAssetCommand::Freeze)
```
JuaniRios: should fix: the idempotency check reads status from the view, then dispatches here, with nothing isolating the two steps. If a concurrent server write changes the asset in that window the aggregate emits zero events, but Store::send still returns Ok, so the CLI prints "Froze SGOV." for what was actually a no-op (the inverse race prints "already frozen" when it isn't). For a manual M1 tool the final state is still safe because the aggregate is the source of truth, so this is a misleading-message bug rather than a correctness one. The clean fix is to have Store::send report whether any events were persisted and derive the outcome from that, which also matters more once the M3 scheduler shares this path.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+/// confirmation is injected so the abort/confirm branches are unit-testable
+/// without driving real stdin. Aborting a mutation returns an error (non-zero
+/// exit) so automation can distinguish "operator declined" from "done".
+async fn execute(
+    admin: &AssetAdmin,
+    action: AssetAction,
+    underlying: &UnderlyingSymbol,
+    confirm: impl Fn(&str) -> io::Result<bool>,
+) -> anyhow::Result<()> {
+    let report = admin.status(underlying).await?.ok_or_else(|| {
```
JuaniRios, line 129: minor: execute reads status here for display, then freeze/unfreeze read current_status again internally to decide idempotency. The operator confirms against the first read but the outcome is decided by the second, so if the state changes while they're reading and typing, the printed status and the reported outcome can disagree. Read the status once and pass the AssetStatusReport into freeze/unfreeze.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+/// fails fast with a clear parse error instead of a later "not a supported
+/// asset" lookup failure, and trims surrounding whitespace so `" SGOV"` matches
+/// `SGOV`.
+fn parse_underlying(value: &str) -> Result<UnderlyingSymbol, String> {
+    let trimmed = value.trim();
+    if trimmed.is_empty() {
+        return Err("underlying symbol must not be empty".to_string());
+    }
+
+    Ok(UnderlyingSymbol::new(trimmed))
```
JuaniRios: minor: this preserves case, but the lookup does an exact view_id match and assets are stored uppercase (SGOV). So issuer freeze sgov reports "not a supported tokenized asset" even though the asset exists, which is an easy way to fumble an ex-date freeze under pressure. Canonicalize with trimmed.to_ascii_uppercase() here and add a mixed-case test.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+    #[error("database error: {0}")]
+    Database(#[from] sqlx::Error),
+    #[error("migration error: {0}")]
+    Migrate(#[from] sqlx::migrate::MigrateError),
+    #[error("failed to read asset view: {0}")]
+    View(#[from] TokenizedAssetViewError),
+    #[error("event store reconcile error: {0}")]
+    Reconcile(#[from] ReconcileError),
+    #[error("failed to dispatch command: {0}")]
+    Dispatch(#[from] Box<AggregateError<LifecycleError<TokenizedAsset>>>),
```
JuaniRios: minor: AGENTS.md wants #[from] variant names to mirror the source type, not the operation. store.load() returns this same AggregateError type, so a future call site would get wrapped in a variant claiming it was a "dispatch". Rename it to Aggregate (and update the #[error] message) to match the other variants in this enum.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+        self.store.send(underlying, command).await.map_err(Box::new)?;
+        Ok(())
+    }
+}
+
+/// Rejects an empty or whitespace-only underlying at the CLI boundary so a typo
+/// fails fast with a clear parse error instead of a later "not a supported
+/// asset" lookup failure, and trims surrounding whitespace so `" SGOV"` matches
+/// `SGOV`.
+fn parse_underlying(value: &str) -> Result<UnderlyingSymbol, String> {
```
JuaniRios, line 340: minor: the trim and reject behavior here isn't tested directly. The only test hits the empty string through clap, so neither the " SGOV" trimming path nor whitespace-only rejection is covered, and dropping the trim would silently change which symbol lands in the event store. Add a unit test calling parse_underlying directly for " SGOV ", "" and "   ".

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+/// without driving real stdin. Aborting a mutation returns an error (non-zero
+/// exit) so automation can distinguish "operator declined" from "done".
+async fn execute(
+    admin: &AssetAdmin,
+    action: AssetAction,
+    underlying: &UnderlyingSymbol,
+    confirm: impl Fn(&str) -> io::Result<bool>,
+) -> anyhow::Result<()> {
+    let report = admin.status(underlying).await?.ok_or_else(|| {
+        anyhow::anyhow!("{underlying} is not a supported tokenized asset")
```
JuaniRios: minor: this not-found error (message plus non-zero exit) is the entry-point behavior for all three subcommands, but no execute test exercises it. The existing not-found test only covers the AssetAdmin methods, not execute, so a refactor of this guard wouldn't be caught. Add an execute test with an unknown symbol that asserts the error message.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+    let mut input = String::new();
+    io::stdin().read_line(&mut input)?;
+
+    Ok(parse_confirmation(&input))
+}
+
+/// Confirmation accepts `y`/`yes` case-insensitively (after trimming);
+/// everything else — including empty input and EOF — declines.
+fn parse_confirmation(input: &str) -> bool {
+    matches!(input.trim().to_ascii_lowercase().as_str(), "y" | "yes")
```
JuaniRios: nit: to_ascii_lowercase allocates a String just to compare against two literals. trimmed.eq_ignore_ascii_case("y") || trimmed.eq_ignore_ascii_case("yes") does the same thing with no allocation and reads a bit clearer.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+    /// Underlying symbol, e.g. SGOV.
+    #[arg(value_parser = parse_underlying)]
+    underlying: UnderlyingSymbol,
+    #[arg(
+        long = "database-url",
+        env = "DATABASE_URL",
+        default_value = DEFAULT_DATABASE_URL,
+        value_parser = parse_sqlite_url
+    )]
+    db: String,
```
JuaniRios: nit: every other reference to this concept is spelled database_url (the flag, the env, Config.database_url, and the sibling field database_max_connections). Renaming db to database_url keeps it consistent and self-explanatory.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+            panic!("expected the freeze subcommand");
+        };
+        assert_eq!(args.underlying, UnderlyingSymbol::new("SGOV"));
+        assert_eq!(args.db, "sqlite:data.db");
+
+        assert!(matches!(
+            IssuerCli::try_parse_from(["issuer", "unfreeze", "SGOV"])
+                .expect("valid args")
+                .command,
+            IssuerCommand::Unfreeze(_)
```
JuaniRios: nit: the unfreeze and status assertions only match the enum variant with a wildcard, so they don't inspect any parsed args and would pass even if the validators were unwired. Either drop them or have them assert the parsed underlying the way the freeze branch does.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+            )
+            .await
+            .expect("Failed to add asset");
+
+        AssetAdmin { store, pool }
+    }
+
+    #[traced_test]
+    #[tokio::test]
+    async fn freeze_then_unfreeze_dispatches_and_is_idempotent() {
```
JuaniRios: nit: this test chains roughly six behaviors (the full lifecycle plus both idempotency paths plus log assertions), so a failure is harder to localize. Consider splitting into a round-trip test and two focused idempotency tests. admin_with_asset gives a fresh store each time so that's cheap to do.

JuaniRios (changes requested):

src/bin/issuer.rs
0xgleb: Fixed. The Dockerfile now patchelfs `target/${BUILD_PROFILE}/issuer` and `COPY`s it into the final image as `./issuer`, so `docker compose exec issuance-bot ./issuer freeze SGOV` resolves.

src/bin/issuer.rs
0xgleb: Fixed. Added `default-run = "st0x-issuance"` under `[package]`, so plain `cargo run` stays unambiguous.

src/tokenized_asset/cli.rs
0xgleb: Fixed. Constrained the arg with `value_parser = clap::value_parser!(u32).range(1..)`, so 0 or an unparsable value is a clean clap error instead of the `ArrayQueue::new(0)` panic.

src/tokenized_asset/cli.rs
0xgleb: Fixed. `execute` now reads the status once and passes it into `freeze`/`unfreeze` (which no longer re-read), so the displayed/confirmed status and the decision cannot disagree.

src/tokenized_asset/cli.rs
0xgleb: Fixed. `parse_underlying` now `to_ascii_uppercase()`s the trimmed symbol, so `issuer freeze sgov` resolves to the stored `SGOV`. Added a mixed-case routing assertion and a `parse_underlying` unit test.

src/tokenized_asset/cli.rs
0xgleb: Fixed. Renamed the variant to `Aggregate` and updated its `#[error]` message to mirror the source type rather than the operation.

src/tokenized_asset/cli.rs
0xgleb: Fixed. Added `parse_underlying_trims_uppercases_and_rejects_blank`, covering trimming/upper-casing, empty, and whitespace-only input.

src/tokenized_asset/cli.rs
0xgleb: Fixed. Added `execute_rejects_unknown_asset`, which drives `execute` with an unknown symbol and asserts the not-a-supported-asset message and non-zero exit.

src/tokenized_asset/cli.rs
0xgleb: Fixed. `parse_confirmation` now uses `eq_ignore_ascii_case` for `y`/`yes` — no allocation.

src/tokenized_asset/cli.rs
0xgleb: Fixed. Renamed the field to `database_url`, matching the flag/env and `Config.database_url`.

src/tokenized_asset/cli.rs
0xgleb: Fixed. The unfreeze/status branches now bind the parsed args and assert the underlying (the unfreeze case feeds lower-case input to also cover the upper-casing).

src/tokenized_asset/cli.rs
0xgleb: Fixed. Split into `freeze_then_unfreeze_round_trip` and `freeze_and_unfreeze_report_idempotent_no_ops`.

src/tokenized_asset/cli.rs
0xgleb: A zero-event no-op is valid CQRS/ES — a command handler produces zero-or-more events, and `Store::send` returning `Ok` on a no-op is by design to support idempotency, so I do not think this warrants changing event-sorcery. The only residual is cosmetic: under a concurrent write landing between the read and the dispatch, the printed line could say "Froze" for what the aggregate treated as a no-op. The final state is still correct (the aggregate is the source of truth), so for a manual M1 admin tool that is acceptable. If we ever need the message to be exact, that is a CLI-only concern — the CLI can re-read state after dispatch and report on the actual transition — rather than a change to the shared event store.

JuaniRios (approved):

PR #183, opened 2026-06-16
feat: dto crate

0xgleb:
## Motivation

The issuance API DTOs were defined inline across the monolith, so neither a
TypeScript dashboard nor external Rust clients (e.g. the liquidity rebalance
guard, RAI-1038) could share the wire types without depending on the whole
backend. Implements RAI-1058.

## Solution

Converts issuance to a Cargo workspace and adds a `st0x-issuance-dto` crate
(`crates/dto`) of shared wire types deriving `ts_rs::TS` + serde. The
tokenized-asset newtypes (`UnderlyingSymbol`, `TokenSymbol`, `Network`) and the
tokenized-asset API DTOs (status/detail/list/add) move there; the main crate
re-exports the newtypes and consumes the DTOs (no duplication). `export_bindings()`
+ a thin binary emit the `.ts` files for a dashboard. Minimal deps
(`alloy-primitives`, not full `alloy`), declared via `[workspace.dependencies]`.

The detail endpoint (`GET /tokenized-assets/<underlying>`) now returns
`status: "enabled" | "frozen"` instead of `enabled: bool`: a bool could not
represent `Frozen`, so a frozen asset was reported as `enabled: true`. The enum
reuses the freeze-status type, mapped from the domain `AssetStatus` via a single
`From` impl. The list/status/add wire formats are unchanged.

Mint/account/redemption DTOs follow in a later increment.

Stacked on the freeze stack (#180/#181/#182).

## Checks

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -433,14 +433,15 @@ Environment variables are defined across the deployment pipeline:
     `Json(#[from] serde_json::Error)`
     - generic, truthful. Use context from where the error is handled.
 - **CRITICAL: Make Invalid States Unrepresentable**: Use ADTs/enums to encode
-  business rules and state transitions directly in types rather than runtime
-  validation.
-  - **FORBIDDEN**: Types with all/most fields `Option`; multiple nullable fields
-    that can contradict; string status fields paired with Options meant to be
-    present only for certain statuses
+  business rules and state transitions in types, not runtime validation.
+  - **FORBIDDEN**: Types with all/most fields `Option`; nullable fields that can
+    contradict; multiple `bool`s encoding one status (e.g.
+    `{ enabled: bool, frozen: bool }` — four combos, only some valid); string
+    status fields paired with Options present only for certain statuses
   - **CORRECT**: Enum variants for mutually exclusive states, with
     state-specific data inside each variant; newtypes to prevent mixing
-    incompatible values
+    incompatible values. A status with N states is ONE enum of N variants, never
+    N booleans (`enum AssetStatus { Enabled, Frozen }`)
   - Example: `enum Account { NotLinked, Linked { client_id, email, ... } }`, NOT
     `struct Account { client_id: Option<ClientId>, email: Option<Email>, ... }`
 - **CRITICAL: Parse, Don't Validate**: If a value exists, it must be valid.
@@ -454,17 +455,11 @@ Environment variables are defined across the deployment pipeline:
   - Applies to ALL constrained domain types: API keys, emails, quantities,
     addresses, IDs, etc.
     ```rust
-    // WRONG: validate() can be forgotten → an invalid value can exist
-    pub struct ApiKey(pub String);
-    impl ApiKey { pub fn validate(&self) -> Result<(), Error> { /* ... */ } }
-
-    // CORRECT: if ApiKey exists, it's guaranteed valid
-    pub struct ApiKey(String); // private inner value
+    // existence ⇒ validity: private inner, fallible new() the only constructor
+    pub struct ApiKey(String);
     impl ApiKey {
         pub fn new(value: String) -> Result<Self, ApiKeyError> {
-            if value.len() < 32 {
-                return Err(ApiKeyError::TooShort { len: value.len() });
-            }
+            if value.len() < 32 { return Err(ApiKeyError::TooShort); }
             Ok(Self(value))
         }
     }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,14 +4,18 @@ version = "0.1.0"
 edition = "2024"
 default-run = "st0x-issuance"
 
+[lints]
+workspace = true
+
 [dependencies]
+st0x-issuance-dto = { path = "crates/dto" }
 event-sorcery = { git = "https://github.com/ST0X-Technology/event-sorcery.git", tag = "0.1.2" }
 sqlite-es = { git = "https://github.com/ST0X-Technology/event-sorcery.git", tag = "0.1.2" }
 cqrs-es = "0.5.0"
 async-trait = "0.1.89"
 rocket = { version = "0.5.1", features = ["json"] }
-serde = { version = "1.0.228", features = ["derive"] }
-serde_json = "1.0.150"
+serde = { workspace = true }
+serde_json = { workspace = true }
 sqlx = { version = "0.9.0", features = [
   "sqlite",
   "chrono",
@@ -24,7 +28,7 @@ thiserror = "2.0.17"
 tracing = "0.1.41"
 chrono = { version = "0.4.42", features = ["serde"] }
 uuid = { version = "1.18.1", features = ["v4", "serde"] }
-clap = { version = "4.5.49", features = ["derive", "env"] }
+clap = { workspace = true, features = ["env"] }
 alloy = { version = "1.0.41", features = [
   "provider-ws",
   "provider-http",
@@ -69,10 +73,20 @@ tempfile = "3.23.0"
 tokio = { version = "1.48.0", features = ["macros"] }
 tracing-test = { version = "0.2.6", features = ["no-env-filter"] }
 
-[lints.rust]
+[workspace]
+members = ["crates/dto"]
+
+[workspace.dependencies]
+alloy-primitives = { version = "1.6.0", features = ["serde"] }
+clap = { version = "4.5.49", features = ["derive"] }
+serde = { version = "1.0.228", features = ["derive"] }
+serde_json = "1.0.150"
+ts-rs = "11.0.1"
+
+[workspace.lints.rust]
 unsafe_code = "forbid"
 
-[lints.clippy]
+[workspace.lints.clippy]
 unwrap_used = "deny"
 enum_glob_use = "allow"
 
--- a/crates/dto/src/lib.rs
+++ b/crates/dto/src/lib.rs
@@ -1,0 +1,422 @@
+//! Wire types for the st0x.issuance HTTP API, shared by the server, Rust
+//! clients, and (via `ts-rs`) the TypeScript dashboard.
+//!
+//! These derive [`ts_rs::TS`] so the dashboard can build against generated
+//! bindings without depending on the backend crate. The serde representation is
+//! the API contract (snake_case) -- do not add `rename_all` without versioning
+//! the endpoints.
+
+use std::path::Path;
+use std::str::FromStr;
+
+use alloy_primitives::Address;
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+/// Underlying equity symbol, e.g. `SGOV`. Also the `TokenizedAsset` aggregate id.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, TS)]
+pub struct UnderlyingSymbol(pub String);
+
+impl UnderlyingSymbol {
+    pub fn new(value: impl Into<String>) -> Self {
+        Self(value.into())
+    }
+}
+
+impl std::fmt::Display for UnderlyingSymbol {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl FromStr for UnderlyingSymbol {
+    type Err = std::convert::Infallible;
+
+    fn from_str(value: &str) -> Result<Self, Self::Err> {
+        Ok(Self::new(value))
+    }
+}
+
+/// Tokenized share symbol, e.g. `tSGOV`.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, TS)]
+pub struct TokenSymbol(pub String);
+
+impl TokenSymbol {
+    pub fn new(value: impl Into<String>) -> Self {
+        Self(value.into())
+    }
+}
+
+impl std::fmt::Display for TokenSymbol {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// Blockchain network the tokenized asset lives on, e.g. `base`.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, TS)]
+pub struct Network(pub String);
+
+impl Network {
+    pub fn new(value: impl Into<String>) -> Self {
+        Self(value.into())
+    }
+}
+
+impl std::fmt::Display for Network {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+/// Single supported asset, as returned by `GET /tokenized-assets/<underlying>`.
+///
+/// Carries the full [`TokenizedAssetStatus`] rather than an `enabled: bool`: a
+/// bool cannot represent `Frozen`, which forced a lossy mapping that reported a
+/// frozen asset as enabled. The enum keeps the freeze state visible to
+/// consumers and the contradictory states unrepresentable.
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+pub struct TokenizedAssetDetailResponse {
+    pub underlying: UnderlyingSymbol,
+    pub token: TokenSymbol,
+    pub network: Network,
+    #[ts(type = "string")]
+    pub vault: Address,
+    pub status: TokenizedAssetStatus,
+}
+
+/// Whether a supported tokenized asset currently accepts new mints.
+///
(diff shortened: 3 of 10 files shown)


0xgleb:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/183?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#189** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/189?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#188** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/188?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#186** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/186?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#185** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/185?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#184** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/184?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#183** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/183?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/183?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#182** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/182?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#181** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/181?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#196** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=183).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

findolor (approved):

src/tokenized_asset/api.rs
```diff
@@ -48,29 +42,49 @@ pub(crate) async fn get_tokenized_asset(
...
-            enabled,
+            status,
             ..
         }) => Ok(Json(TokenizedAssetDetailResponse {
             underlying,
             token,
             network,
             vault,
-            enabled,
+            enabled: status.is_listed(),
```
JuaniRios: should fix: `enabled: status.is_listed()` returns true for both Enabled and Frozen, so a frozen asset reports `enabled: true` on `GET /tokenized-assets/<underlying>`. That silently flips the field's meaning from "accepts mints" to "is listed," and it ships straight into the TS bindings. Use `enabled: matches!(status, AssetStatus::Enabled)` (or `!status.is_frozen()`) so a frozen asset reports false.

src/tokenized_asset/api.rs
```diff
@@ -48,29 +42,49 @@ pub(crate) async fn get_tokenized_asset(
             token,
             network,
             vault,
-            enabled,
+            status,
```
JuaniRios, line 45: should fix: nothing tests this detail handler with a frozen asset. The only `enabled` coverage hardcodes `enabled: true` directly on the DTO struct, which bypasses the `is_listed()` mapping that's actually wrong here. Add an integration test that freezes an asset, GETs the detail endpoint, and asserts `enabled: false`, mirroring the existing status-endpoint tests.

src/tokenized_asset/view.rs
```diff
@@ -19,26 +23,34 @@ pub(crate) enum TokenizedAssetViewError {
...
+/// SPEC.md).
 pub(crate) async fn list_enabled_assets(
     pool: &Pool<Sqlite>,
 ) -> Result<Vec<TokenizedAssetView>, TokenizedAssetViewError> {
     let rows = sqlx::query!(
         r#"
         SELECT json_extract(payload, '$.Live') as "live: String"
         FROM tokenized_asset_view
-        WHERE json_extract(payload, '$.Live.enabled') = 1
+        WHERE json_extract(payload, '$.Live.status') IN ('Enabled', 'Frozen')
```
JuaniRios, line 53: should fix: this `WHERE json_extract(... '$.Live.status') IN (...)` silently drops any row whose `$.Live` is NULL (non-live or corrupt projection). `load_asset_by_underlying` fails fast with `NonLiveRow` on that same condition, so the two helpers disagree. Since this feeds redemption monitoring and receipt backfill, a corrupt row should surface as an error, not quietly disappear from the list.

crates/dto/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+
+/// Single supported asset, as returned by `GET /tokenized-assets/<underlying>`.
+#[derive(Debug, Clone, Serialize, Deserialize, TS)]
+pub struct TokenizedAssetDetailResponse {
+    pub underlying: UnderlyingSymbol,
+    pub token: TokenSymbol,
+    pub network: Network,
+    #[ts(type = "string")]
+    pub vault: Address,
+    pub enabled: bool,
```
JuaniRios: should fix: `enabled: bool` can't represent the Frozen state even though `TokenizedAssetStatus` already models Enabled/Frozen/Unavailable. Carrying a bool is what forced the lossy `is_listed()` mapping in the handler. Consider using the status enum on the detail response so consumers see the real state.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+    ) -> Result<Self, AssetAdminError> {
+        let connect_options = SqliteConnectOptions::from_str(db)?
+            .busy_timeout(Duration::from_secs(5));
+
+        let pool = SqlitePoolOptions::new()
+            .max_connections(max_connections)
+            .connect_with(connect_options)
+            .await?;
+
+        sqlx::migrate!("./migrations").run(&pool).await?;
```
JuaniRios, line 252: should fix: the read-only `status` subcommand runs `sqlx::migrate!().run()` before doing anything. A status query shouldn't mutate the schema, and it'll fail or block if the operator only has read access. Gate migrations to the commands that actually need a writable, current schema.

src/tokenized_asset/view.rs
JuaniRios, line 53: should fix: AGENTS.md says to read views via `GenericQuery::load()` instead of parsing JSON with raw SQL, and both new helpers hand-roll `json_extract` against the payload. For a cross-row query a dedicated read model with real columns (or iterating IDs via `GenericQuery`) keeps this within the documented pattern.

crates/dto/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+use std::path::Path;
+use std::str::FromStr;
+
+use alloy_primitives::Address;
+use serde::{Deserialize, Serialize};
+use ts_rs::TS;
+
+/// Underlying equity symbol, e.g. `SGOV`. Also the `TokenizedAsset` aggregate id.
+#[derive(Debug, Clone, PartialEq, Eq, Hash, Serialize, Deserialize, TS)]
+pub struct UnderlyingSymbol(pub String);
```
JuaniRios, line 18: minor: the newtypes expose their inner value as `pub String` (`UnderlyingSymbol(pub String)`, etc.), so callers can construct one bypassing `new()`. The parse-don't-validate rule wants the inner private with the smart constructor as the only path. Worth locking down before this public crate's API ossifies, even if `new()` does little validation today.

AGENTS.md
```diff
@@ -130,11 +130,11 @@ audit trail, enable time-travel debugging, and provide a single source of truth.
 - `sqlx migrate revert` - Revert last migration
 - `sqlx db reset -y` - Drop the database and re-run all migrations
 - Database URL configured via `DATABASE_URL` environment variable
-- **Fixing "unable to open database file" errors** - If `cargo build` fails with
-  sqlx macro errors like
-  `error returned from database: (code: 14) unable to
-  open database file`, run
-  `sqlx db reset -y` to recreate the database
+- **CRITICAL: For ANY sqlx macro / compile-time DB error, the fix is
```
JuaniRios, line 133: minor: AGENTS.md is 40,275 chars at this commit, just over the 40,000 limit the file sets for itself. Worth trimming a little to stay under.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+            .await?
+            .map(|view| view.status))
+    }
+
+    async fn dispatch(
+        &self,
+        underlying: &UnderlyingSymbol,
+        command: TokenizedAssetCommand,
+    ) -> Result<(), AssetAdminError> {
+        self.store.send(underlying, command).await.map_err(Box::new)?;
```
JuaniRios: minor: `.map_err(Box::new)?` hand-rolls a conversion that a `#[from]` on the error variant would give you via plain `?`. Adding `#[from]` for the dispatch error keeps it idiomatic and matches the error-handling guidance.

src/tokenized_asset/api.rs
```diff
@@ -48,29 +42,49 @@ pub(crate) async fn get_tokenized_asset(
...
+            .await
+            .map_err(|err| {
+                error!(target: "asset", error = %err,
+                    "Failed to load tokenized asset status"
+                );
+                Status::InternalServerError
+            })?;
+
+    match view {
+        Some(TokenizedAssetView { underlying, status, .. }) => {
```
JuaniRios, line 78: minor: this maps `AssetStatus` to the response enum arm-by-arm inline. A `From<AssetStatus>` impl would centralize the mapping and keep the handler readable, especially since another endpoint does a similar conversion.

src/tokenized_asset/mod.rs
```diff
@@ -10,62 +11,43 @@ use event_sorcery::{EventSourced, Never, Table};
...
 }
 
-impl UnderlyingSymbol {
-    pub fn new(value: impl Into<String>) -> Self {
-        Self(value.into())
+impl AssetStatus {
+    /// Whether the asset is supported (appears in `list_enabled_assets`). Both
+    /// `Enabled` and `Frozen` assets are supported — freezing gates minting, not
+    /// listing, so this must never conflate freeze with de-listing.
+    pub(crate) const fn is_listed(&self) -> bool {
```
JuaniRios: minor: `is_listed()` returns true for both Enabled and Frozen. Once the `enabled` bug in the detail handler is fixed, check whether it still has any callers. If it stays, it needs a test pinning the both-states-true behavior, since that invariant is load-bearing for the freeze design and currently untested.

src/mint/api/initiate.rs
```diff
@@ -220,6 +222,200 @@ mod tests {
...
+
+        assert!(logs_contain_at!(
+            tracing::Level::WARN,
+            &["Rejecting mint for frozen asset"]
+        ));
+    }
+
+    #[traced_test]
+    #[tokio::test]
+    async fn test_initiate_mint_accepts_unfrozen_asset() {
```
JuaniRios, line 317: minor: this test asserts both that a frozen asset is rejected and that an unfrozen one succeeds in one function. Splitting into two makes a failure point directly at which behavior regressed.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+            );
+        }
+
+        for decline in ["", "n", "N", "no", "yep", "  ", "\n"] {
+            assert!(!parse_confirmation(decline), "{decline:?} should decline");
+        }
+    }
+
+    #[test]
+    fn issuer_cli_routes_subcommands_and_honors_explicit_db() {
```
JuaniRios: minor: this test only verifies Clap argument routing, not business logic, so it's mostly exercising the framework. Consider dropping it or folding the routing check into a test that also exercises the command's effect.

src/tokenized_asset/cli.rs
```diff
@@ -0,0 +1,614 @@
...
+            |view| AssetStatusReport {
+                underlying: underlying.clone(),
+                status: view.status,
+            },
+        ))
+    }
+
+    /// Freezes the asset. Idempotent — a frozen asset reports `AlreadyFrozen`
+    /// without emitting another event; an unknown asset reports `NotFound`.
+    pub(crate) async fn freeze(
```
JuaniRios, line 283: minor: freeze/unfreeze reads the view status, dispatches the command, then reports an outcome based on the earlier read. If state changes between the read and the dispatch, the reported outcome can be misleading. Reporting from the command result rather than the pre-read closes the TOCTOU gap.

src/tokenized_asset/mod.rs
```diff
@@ -10,62 +11,43 @@ use event_sorcery::{EventSourced, Never, Table};
...
+pub(crate) use st0x_issuance_dto::{Network, TokenSymbol};
+
+/// Whether an asset accepts new mints.
+///
+/// `Frozen` gates *only* new minting — a frozen asset stays supported and in the
+/// `list_enabled_assets()` set so in-flight redemptions still detect and
+/// complete. This is orthogonal to listing; see the freeze invariant in
+/// SPEC.md. Serializes to the bare strings `"Enabled"` / `"Frozen"`, which the
+/// view queries match against `$.Live.status`.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
JuaniRios: nit: `AssetStatus` is a field-less enum but doesn't derive `Copy`, so it gets moved/cloned where a copy would do. Adding `Copy` is cheap and more ergonomic.

src/tokenized_asset/mod.rs
```diff
@@ -128,73 +114,153 @@ impl EventSourced for TokenizedAsset {
...
-            added_at: Utc::now(),
-        }])
+        match command {
+            TokenizedAssetCommand::Add {
+                underlying,
+                token,
+                network,
+                vault,
+            } => {
+                tracing::info!(target: "asset", underlying = %underlying.0,
```
JuaniRios: nit: these tracing calls log `%underlying.0`, reaching past the newtype into the inner string instead of using its `Display`. Logging `%underlying` keeps formatting in one place.

Cargo.toml
```diff
@@ -68,10 +72,20 @@ tempfile = "3.23.0"
...
-[lints.rust]
+[workspace]
+members = ["crates/dto"]
+
+[workspace.dependencies]
+alloy-primitives = { version = "1.6.0", features = ["serde"] }
+clap = { version = "4.5.49", features = ["derive"] }
+serde = { version = "1.0.228", features = ["derive"] }
+serde_json = "1.0.150"
+ts-rs = "11.0.1"
```
JuaniRios, line 84: nit: `ts-rs = "11.0.1"` resolves to 11.1.0 via the caret range, so the lockfile drives the wire-format guarantee, not this pin. If the exact ts-rs version matters for the emitted bindings, pin it precisely.

src/tokenized_asset/mod.rs
```diff
@@ -315,6 +381,163 @@ mod tests {
...
+            .await
+            .then_expect_events(&[]);
+
+        assert!(logs_contain_at!(
+            tracing::Level::WARN,
+            &["Ignoring freeze/unfreeze for an asset that does not exist"]
+        ));
+    }
+
+    #[test]
```
JuaniRios, line 521: nit: test_apply_frozen / test_apply_unfrozen look like they duplicate coverage already exercised by higher-level tests, with weaker assertions. Not a blocker, just flagging the redundancy.

crates/dto/src/lib.rs
```diff
@@ -0,0 +1,365 @@
...
+            TokenSymbol::new("tSGOV")
+        );
+        assert_eq!(
+            serde_json::from_value::<Network>(json!("base")).unwrap(),
+            Network::new("base")
+        );
+    }
+
+    #[test]
+    fn newtypes_display_their_inner_value() {
```
JuaniRios, line 207: nit: this asserts a derived, trivial Display impl, so it's low value. Optional to drop.

JuaniRios (changes requested):

crates/dto/src/lib.rs
0xgleb: Holding off here -- a String newtype with a no-op new() carries no invariant, so making the field private just adds ceremony. Proper modeling is in #188 (Network -> enum); UnderlyingSymbol/TokenSymbol get validating/derived treatment in a follow-up.

JuaniRios (approved):

PR #204, opened 2026-06-24
NixOS stack deployment setup

rouzwelt:
## Motivation

This PR migrates the issuance service deployment from Docker-based infrastructure to a declarative NixOS stack, mirroring the approach successfully deployed for `st0x.liquidity`. The existing deployment relies on manually-managed Docker Compose workflows that lack reproducibility, atomic rollbacks, and proper secrets managemen.

Adopting the same NixOS stack used for liquidity unifies operational patterns across both services and provides the foundation for reliable production operations.

## Solution

This PR introduces a complete NixOS-based deployment pipeline with the following components:

**Infrastructure layer (Terraform + NixOS)**

- Terraform modules in `infra/` provision DigitalOcean droplets, block storage, reserved IPs, and firewall rules for both prod and staging
- `disko.nix` defines declarative disk partitioning
- `bootstrap.nix` handles initial NixOS installation via `nixos-anywhere`
- `os.nix` provides the full system configuration including Tailscale integration, fail2ban, and service units

**Deployment orchestration (deploy-rs)**

- `deploy.nix` defines per-environment nodes and per-service profiles
- System profile handles OS-level changes (users, firewall, packages)
- Service profiles enable fast replace (near zero-downtime) binary upgrades without full system rebuilds
- Explicit activation ordering via `services.nix` ensures coordinated restarts
- Automatic rollback on failed deployments

**Secrets management (agenix)**

- `secret/secrets.nix` defines encryption rules for env files and Fireblocks keys
- `keys.nix` centralizes SSH public keys for humans, CI, and host identities
- `scripts/secret.nu` provides an ergonomic CLI for editing encrypted secrets with automatic rekeying
- Per-environment Tailscale auth keys enable secure node enrollment

**Build pipeline**

- `nix/mk-abi.nix` and `nix/abis.nix`replace the `lib/ethgild` submodule with a Nix-managed Forge build
- `rust.nix` packages the issuance server and operator CLI (issuer) with ABI paths injected as compile-time env vars
- Removes `prepSolArtifacts` step from CI (ABIs are now pure Nix derivations)

**Networking (Tailscale)**

- `nix/tailscale.nix` enrolls each host onto the tailnet and marks `tailscale0` as a trusted firewall interface
- SSH and the issuance API are accessible **only**over the tailnet (no public exposure)
- DigitalOcean firewall permits only Tailscale WireGuard traffic (UDP 41641)

**GitHub Actions workflows**

- `deploy-prod.yaml`: deploys tagged releases to production with tag-on-main validation
- `deploy-staging.yaml`: deploys arbitrary branches/commits to staging for testing
- `rekey.yaml`: rotates secret encryption when host keys or recipient lists change
- All workflows resolve deploy targets via Tailscale MagicDNS and use SSH key verification

**Developer tooling**

- `nix run .#bootstrap` installs NixOS on a fresh droplet and captures its host key
- `nix run .#prodDeployAll` / `stagingDeployAll` deploy all profiles to the respective environment
- `nix run .#secret <file>` edits encrypted secrets with 1Password / SSH identity support
- `nix run .#rekey` rotates all secrets after key changes

**Migration path**

- `.env.secret.example` documents the expected secret structure (replaces `.env.example` for production)
- Existing Docker-based workflows have been removed
- Dockerfile and docker-compose file and template along side `rollback.sh` files have been removed

This change does **not** alter application behavior — the issuance server runs the same binary with the same env vars and database. It only changes **how**that binary is built, deployed, and configured

## Checks

- [x] linked any relevant issues or PRs https://linear.app/makeitrain/issue/RAI-635/migrate-issuance-deployment-to-nixos-stack-mirror-st0xliquidity

## Summary by CodeRabbit

* **New Features**
  * Added manually-triggered production and staging deployment workflows.
  * Introduced automated host bootstrapping and a secrets rekey workflow.
  * Enhanced encrypted environment/secret handling and added Tailscale-based connectivity for deployed hosts.
  * Added infrastructure provisioning and deployment helper tooling backed by Terraform.
* **Bug Fixes**
  * Improved deployment reliability with SSH known-host handling and connectivity warm-up.
* **Documentation**
  * Updated example environment files, including a populated secrets template and an emptied general template.
* **Chores**
  * Removed older Docker/rollback-based deployment workflows and related scripts.


scripts/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+
+valid_key='^ssh-ed25519 [A-Za-z0-9+/=_]+$'
+if [ -z "$new_key" ] || ! echo "$new_key" | grep -qE "$valid_key"; then
+  echo "ERROR: SSH host key is empty or malformed: '$new_key'" >&2
+  exit 1
+fi
+
+sed -i \
+  "/$host_key_field =/{n;s|\"ssh-ed25519 [A-Za-z0-9+/=_]*\"|\"$new_key\"|;}" \
+  keys.nix
```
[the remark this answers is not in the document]
rouzwelt: don't think that's the case though

flake.nix
```diff
@@ -112,12 +292,10 @@
...
           // {
-            prepSolArtifacts = rainix.mkTask.${system} {
+            prepSolArtifacts = pkgs.writeShellApplication {
               name = "prep-sol-artifacts";
-              additionalBuildInputs = rainix.sol-build-inputs.${system};
-              body = ''
-                set -euxo pipefail
-                (cd lib/ethgild && forge build)
+              text = ''
+                ln -sfn ${abis.ethgild}/out abis
```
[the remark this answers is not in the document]
rouzwelt: not needed

rust.nix
```diff
@@ -0,0 +1,120 @@
...
+
+  # Overlay the forge-built ABIs over the cleaned source so the `sol!()`
+  # macros in src/bindings.rs can resolve them.
+  fullSrc = pkgs.runCommand "st0x-issuance-src" { } ''
+    cp -rL --no-preserve=mode ${baseSrc} $out
+    chmod -R u+w $out
+    rm -rf $out/lib/ethgild/out
+    mkdir -p $out/lib/ethgild
+    cp -r ${ethgildAbis}/out $out/lib/ethgild/out
+  '';
```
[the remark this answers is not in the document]
rouzwelt: fixed

infra/modules/stack/main.tf
```diff
@@ -0,0 +1,83 @@
...
+resource "digitalocean_reserved_ip_assignment" "nixos" {
+  ip_address = digitalocean_reserved_ip.nixos.ip_address
+  droplet_id = digitalocean_droplet.nixos.id
+}
+
+resource "digitalocean_firewall" "st0x_issuance" {
+  name        = "st0x-issuance-${var.environment}"
+  droplet_ids = [digitalocean_droplet.nixos.id]
+
+  # Tailscale WireGuard -- must be publicly reachable for NAT traversal.
```
JuaniRios, line 57: critical: this firewall only allows inbound UDP/41641, copied verbatim from liquidity. That works for liquidity because its HTTP surface is tailnet-internal, but issuance has to receive Alpaca's inbound callbacks (/accounts/connect, /inkind/issuance, /inkind/issuance/confirm) from an external party that can't join our tailnet, so as written every callback is dropped at the edge and the mint/redeem flow is unreachable. Add a public inbound TCP rule (443) fronted by a TLS proxy, ideally restricted to Alpaca's IP ranges. Note os.nix allowedTCPPorts=[] blocks it at the host layer too, so both layers need fixing.

rust.nix
```diff
@@ -0,0 +1,108 @@
...
+      (craneLib.filterCargoSources path type) || (pkgs.lib.hasInfix "/migrations" path);
+  };
+
+  fullSrc = baseSrc;
+
+  # Git dependency output hashes — update when Cargo.lock is re-pinned.
+  baseVendorDir = craneLib.vendorCargoDeps {
+    src = fullSrc;
+    outputHashes = {
+      "git+https://github.com/ST0X-Technology/event-sorcery.git?tag=0.1.1#d0ad3f66bb5a1da23161d898027bff1abb9dabd6" =
```
JuaniRios: critical: this event-sorcery outputHashes key is tag=0.1.1#d0ad3f6..., but Cargo.lock at this commit pins 0.1.2#8f5c81f.... vendorCargoDeps looks up by the exact Cargo.lock source string, so this key never matches and nix aborts at evaluation. The same stale key is also in flake.nix around line 181, so flake check, every package build, and every deploy fail before building anything. Update both keys to the 0.1.2 source and recompute the sha256, matching liquidity's rust.nix which mirrors the lock URL exactly.

infra/main.tf
```diff
@@ -0,0 +1,20 @@
+module "prod" {
+  source = "./modules/stack"
+
+  environment        = "prod"
+  do_token           = var.do_token
+  droplet_size       = var.prod_droplet_size
+  droplet_name       = "st0x-issuance-nixos"
```
JuaniRios, line 33: critical: this provisions brand new volumes and a new reserved IP, but there's no committed tfstate and no moved/import blocks, so a fresh terraform apply won't adopt the existing prod droplet, volume, or IP. The result is an empty SQLite event store (losing all mint/redemption/account history) plus a new IP that Alpaca isn't pointed at, orphaning the running host. Liquidity handles this with moved{} blocks and a committed terraform.tfstate.age. Add an adoption/import path and a written cutover runbook before this can deploy.

keys.nix
```diff
@@ -0,0 +1,69 @@
...
+    # purpose: deployments from github actions
+    ci = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIARWUchVuQvkFm2tzspdC79hhRyYbWzRjs5iimhxewUy";
+    # purpose: initial key for Digital Ocean to include when first creating
+    # the instance and it's matched by name in their dashboard
+    st0x-op = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIPZ56nOYbGDd0ZfbqxeY7AbvaQGQrHnlC80ccpRGpCoj";
+
+    # purpose: used by the remote host to decrypt ragenix secrets we need for
+    # the system to work at all. auto-generated during bootstrap when
+    # nixos-anywhere turns the initially ubuntu instance into nixos.
+    # PLACEHOLDER: replace with real key after `nix run .#prodBootstrap` (see BOOTSTRAP.md).
```
JuaniRios: critical: these host public keys look like all-zero placeholders, and the PR commits zero .age files even though the config references secret/tailscale-authkey-*.age, st0x-issuance.env.age, and the fireblocks key. Liquidity commits real host keys and 8 .age files. As-is nix evaluation fails on the missing paths, and even once they exist they'd be encrypted to placeholder host keys so the host can't decrypt them until a bootstrap, rekey, redeploy cycle. Capture the real host keys and commit the encrypted secrets.

deploy.nix
```diff
@@ -0,0 +1,258 @@
...
+      fireblocksKeyFile = ./secret/${cfg.encryptedFireblocksKey};
+    in
+    activate.custom pkg (
+      builtins.concatStringsSep " && " [
+        "systemctl stop ${name} || true"
+        "rm -f ${cfg.markerFile}"
+        "mkdir -p /run/st0x /run/agenix"
+
+        # Decrypt env file — contains all secret env vars
+        "${rage} -d -i ${hostKey} ${envSecretsFile} | install -D -m 0640 -o root -g st0x /dev/stdin ${cfg.decryptedEnvPath}"
```
JuaniRios, line 75: should fix: this rage | install pipeline has no set -o pipefail, so if rage fails the pipeline takes install's exit code (0), writes an empty credential file, and the && chain restarts the service with empty creds. Liquidity survives the same pipe because it runs a validate-config gate right after decryption, which issuance dropped, and there's a second unguarded pipe here for the fireblocks key. Add set -o pipefail as the first activation step and restore a post-decrypt validation check before restart.

secret/secrets.nix
```diff
@@ -0,0 +1,55 @@
...
+  ) [ ] issuanceServiceNames;
+
+  envSecretRules = builtins.listToAttrs (
+    map (name: {
+      name = services.${name}.encryptedEnvSecret;
+      value.publicKeys = allServiceKeys;
+    }) issuanceServiceNames
+  );
+
+  fireblocksKeyRules = builtins.listToAttrs (
```
JuaniRios, line 38: should fix: the fireblocks RSA signing key is encrypted to both the prod and staging host keys, so a staging host compromise unlocks the production blockchain signer. Liquidity only shared a TOML config this way, not a signing key, so the blast radius here is much larger. Encrypt the prod fireblocks key only to the prod host and staging only to staging, with per-environment recipient sets.

.github/workflows/deploy-prod.yaml
```diff
@@ -0,0 +1,91 @@
...
+jobs:
+  deploy:
+    environment: production
+    concurrency:
+      group: deploy-prod
+      cancel-in-progress: false
+    runs-on: ubuntu-latest
+    steps:
+      - name: Free disk space
+        uses: jlumbroso/free-disk-space@v1.3.1
```
JuaniRios: should fix: the third-party actions in these new workflows are pinned to floating tags rather than commit SHAs, which the deleted workflows pinned properly, and rekey.yaml runs ragenix from a live GitHub flake URL that bypasses flake.lock. Pin every third-party action to a full commit SHA and pin ragenix to the locked flake input.

.github/workflows/rekey.yaml
```diff
@@ -0,0 +1,69 @@
...
+        continue-on-error: true
+
+      - name: Setup SSH key
+        run: |
+          mkdir -p ~/.ssh
+          chmod 700 ~/.ssh
+          echo "${{ secrets.SSH_KEY }}" > ~/.ssh/id_ed25519
+          chmod 600 ~/.ssh/id_ed25519
+
+      - name: Rekey agenix secrets
```
JuaniRios: should fix: this rekey uses the CI identity, but the CI key isn't a recipient of the service secrets it needs to decrypt and rekey, so it can't actually rekey them. It also only git-adds secret/*.age and never infra/*.age, leaving infra secrets without a rekey path. Either add the CI key to the relevant recipient sets or scope this workflow to the secrets CI can decrypt, and include infra/*.age in the add.

flake.nix
```diff
@@ -132,31 +305,105 @@
...
+                      ${env})
+                        flake_config="${environments.${env}.nodeName}-bootstrap"
+                        host_key_field="host-${env}" ;;'') envNames
+                  )}
+                  *)
+                    echo "ERROR: unknown environment '$env'" >&2
+                    exit 1 ;;
+                esac
+
+                ${infraPkgs.parseIdentity}
```
JuaniRios, line 323: should fix: the parsed 1Password identity is written to a tmpfile but there's no trap to remove it before the exec, so the decrypted identity lingers on disk. Add trap _cleanup_identity EXIT right after parseIdentity and before the exec, in both the bootstrap and rekey wrappers.

os.nix
```diff
@@ -0,0 +1,173 @@
...
+    users.root.openssh.authorizedKeys.keys = envRoles.ssh;
+    users.st0x = {
+      isSystemUser = true;
+      group = "st0x";
+    };
+    groups.st0x = { };
+  };
+
+  networking.firewall = {
+    enable = true;
```
JuaniRios, line 120: should fix: services.openssh.openFirewall defaults to true, which opens TCP 22 on the public interface at the NixOS layer, so public SSH is blocked only by the DO cloud firewall rather than defense-in-depth. Set services.openssh.openFirewall = false and rely on the trusted tailscale0 interface for tailnet SSH.

infra/default.nix
```diff
@@ -0,0 +1,398 @@
...
+  '';
+
+  preamble = ''
+    ${parseIdentity}
+    on_exit() { ${cleanup}; _cleanup_identity; }
+    trap on_exit EXIT
+    ${vars.decrypt}
+  '';
+
+  preambleWithEncrypt = ''
```
JuaniRios, line 152: minor: _cleanup_identity is skipped if state.encrypt fails inside the on_exit trap while set -e is active, leaving the decrypted identity behind. Add || true after the non-critical steps in on_exit so cleanup always runs.

deploy.nix
```diff
@@ -0,0 +1,258 @@
...
+  # If any step fails, deploy-rs exits non-zero and rolls back automatically.
+  mkIssuanceProfile =
+    _env: name:
+    let
+      cfg = enabled.${name};
+      pkg = self.packages.${system}.${cfg.package};
+      envSecretsFile = ./secret/${cfg.encryptedEnvSecret};
+      fireblocksKeyFile = ./secret/${cfg.encryptedFireblocksKey};
+    in
+    activate.custom pkg (
```
JuaniRios, line 64: minor: this activation does an explicit systemctl stop then restart (plus secret decrypt), so there's a real if small downtime window, but the PR description calls it zero-downtime. Either drop the explicit stop and rely on restart, or describe it as a fast in-place restart.

.github/workflows/deploy-prod.yaml
```diff
@@ -0,0 +1,91 @@
...
+              exit 0
+            fi
+            echo "Tailscale ping failed on attempt $attempt; retrying..."
+            sleep 5
+          done
+          tailscale status
+          exit 1
+
+      - name: Deploy all profiles
+        run: nix run .#prodDeployAll
```
JuaniRios, line 87: minor: liquidity's deploy workflows have a NixOS-only boot-and-verify step that these don't. Consider adding the same post-deploy health check so a bad activation is caught here rather than in prod.

infra/backend.tf
```diff
@@ -0,0 +1,14 @@
+terraform {
+  required_version = ">= 1.5"
+
+  required_providers {
```
JuaniRios, line 4: minor: infra/.terraform.lock.hcl isn't committed, so provider versions and checksums aren't pinned the way liquidity pins them. Commit the lock file so terraform runs are reproducible.

.env.secrets.example
```diff
@@ -0,0 +1,29 @@
+# Template for the st0x-issuance.env.age secret.
+# DO NOT commit a populated version of this file — it contains credentials.
+#
+# To create the encrypted secret:
+#   1. Copy this file: cp example.secrets.env /tmp/st0x-issuance.env
```
JuaniRios: minor: step 1 here references a filename that doesn't exist, so the cp command fails with no such file, and the ragenix flag shown is invalid. Fix the filename and the flag so the documented setup actually runs.

nix/upgradeable-services.nix
```diff
@@ -0,0 +1,109 @@
...
+      ]
+    else
+      [ ];
+
+  mkService = name: cfg: {
+    description = cfg.description or "st0x ${name}";
+
+    # Service is started by deploy.nix profile, not by systemd on boot.
+    # This avoids coordination issues during deployments.
+    wantedBy = [ ];
```
JuaniRios, line 55: nit: wantedBy is empty and the ready marker lives on tmpfs, so the service won't auto-start after a reboot. This is identical to liquidity and is an intentional shared design (deploy.nix starts the service, not systemd on boot), so it's not a regression in this PR. If reboot auto-start is wanted for a 24/7 service, fix it in both repos rather than here alone.

deploy.nix
```diff
@@ -0,0 +1,258 @@
...
+      infraPkgs,
+      localSystem,
+    }:
+    let
+      deployInputs = infraPkgs.buildInputs ++ [
+        deploy-rs.packages.${localSystem}.deploy-rs
+        pkgs.openssh
+      ];
+
+      deployFlags =
```
JuaniRios, line 192: nit: --skip-checks is hardcoded in deployFlags, which bypasses deploy-rs pre-flight checks. This is copied verbatim from liquidity and CI runs nix flake check separately, so it's inherited convention, not a new issue. Flagging for awareness only; change both repos together if you want checks on deploy.

deploy.nix
0xgleb, line 75: critical: the issuance env file and Fireblocks key are decrypted only here during deploy-rs activation, landing in tmpfs /run/agenix. Tailscale keys go through age.secrets so they survive a reboot, but these do not, and after any reboot the unit stays down for three reasons at once: wantedBy = [], the /run/st0x/*.ready marker is gone, and the EnvironmentFile/Fireblocks paths no longer exist. Worth managing these through age.secrets in os.nix like the Tailscale key, or a oneshot that decrypts before the service starts.

.github/workflows/deploy-staging.yaml
```diff
@@ -0,0 +1,78 @@
+name: Deploy to Staging
+
+on:
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Branch or commit to deploy"
```
0xgleb, line 7: should fix: this accepts any ref, checks it out, and runs nix run .#stagingDeployAll with the deploy SSH key and TS_AUTHKEY, and it pins host keys from the checked-out keys.nix. Unlike deploy-prod.yaml there is no tag-on-main guard, so a malicious branch can run arbitrary build/deploy logic with staging deploy secrets and pin its own host key. Splitting build (untrusted ref, no secrets) from deploy (pinned trusted ref) and allowlisting refs would close this.

secret/secrets.nix
```diff
@@ -0,0 +1,55 @@
...
+  services =
+    (import ../services.nix {
+      lib = {
+        filterAttrs = _: attrs: attrs;
+      };
+    }).byName;
+
+  # Deduplicate keys across environments (st0x-op appears in both roles).
+  dedup = builtins.foldl' (acc: key: if builtins.elem key acc then acc else acc ++ [ key ]) [ ];
+  allServiceKeys = dedup (roles.prod.service ++ roles.staging.service);
```
0xgleb: should fix: allServiceKeys unions prod and staging service roles, so both st0x-issuance.env.age and the Fireblocks key are encrypted to host-staging as well as host-prod. A staging host compromise then decrypts production Alpaca credentials, RPC keys, and the Fireblocks identity. The tailscale keys just below are correctly scoped per environment, so these should be too: separate prod/staging files scoped to each env's roles.<env>.service.

rust.nix
0xgleb: should fix: this pins event-sorcery 0.1.1 (d0ad3f66...) but Cargo.lock is on 0.1.2 (8f5c81f3...), so crane's vendorCargoDeps will fail on a hash mismatch when the real binary is built. flake.nix was already fixed to 0.1.2 in 2588b51; this new file reintroduces the old hash. Copy the 0.1.2 key/hash over and run nix build .#st0x-issuance to confirm.

scripts/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+
+set -euo pipefail
+
+: "${env:?env not set by wrapper}"
+: "${flake_config:?flake_config not set by wrapper}"
+: "${host_key_field:?host_key_field not set by wrapper}"
+: "${identity:?identity not set by wrapper}"
+
+# Resolve IP from terraform state.
+trap "rm -f infra/terraform.tfstate" EXIT
```
0xgleb: should fix: this EXIT trap plus the later explicit rm -f infra/terraform.tfstate both fire unconditionally, and neither checks whether the script created the file. In the plaintext-only case (before the first encrypt) that deletes the operator's only copy of Terraform state, orphaning the droplet/volume/firewall in DO. Only delete plaintext that this script's decrypt step created, or re-encrypt before removing.

nix/upgradeable-services.nix
```diff
@@ -0,0 +1,109 @@
...
+  # Secrets are supplied via EnvironmentFile (issuance kind) or
+  # left entirely absent (plain kind).
+  staticEnvironment =
+    _name: cfg:
+    if cfg.kind == "st0x" then
+      [
+        "DATABASE_URL=sqlite:///mnt/data/issuance.db"
+        "FIREBLOCKS_SECRET_PATH=${cfg.decryptedFireblocksKeyPath}"
+        "ALPACA_API_BASE_URL=https://broker-api.alpaca.markets"
+        # Gates dev-only surfaces (OpenAPI docs). "production" hides them;
```
0xgleb, line 35: should fix: ALPACA_API_BASE_URL is hardcoded to the live broker for every environment, and because services.nix keys the env file by name and mkIssuanceProfile ignores _env, staging decrypts the same env file and Fireblocks key as prod. So staging boots with production credentials pointed at the live Alpaca broker and the prod Fireblocks signing key, and deploy-staging.yaml lets any branch deploy there. For financial infra that means a staging test can journal real shares and trigger real on-chain mints. Staging needs its own secrets, a sandbox Alpaca endpoint, and a staging Fireblocks identity.

scripts/bootstrap.sh
```diff
@@ -0,0 +1,81 @@
...
+host_ip=$(jq -r ".outputs.${env}_droplet_ipv4.value" infra/terraform.tfstate)
+rm -f infra/terraform.tfstate
+
+if [ -z "$host_ip" ] || [ "$host_ip" = "null" ]; then
+  echo "ERROR: could not resolve IP from terraform output '${env}_droplet_ipv4'" >&2
+  exit 1
+fi
+
+ssh_opts=(
+  -o StrictHostKeyChecking=no
```
0xgleb, line 43: should fix: the post-reboot host-key grab uses StrictHostKeyChecking=no over the public IP, then writes the returned key into keys.nix and immediately rekeys every secret to it. A MITM during that window can return their own key (the only check is a format regex) and then decrypt all rekeyed secrets, including the Fireblocks private key. Grabbing the host key from the nixos-anywhere kexec context where it is already known, or verifying the same key over a second independent path before writing it, avoids this.

.github/workflows/deploy-prod.yaml
```diff
@@ -0,0 +1,91 @@
...
+          extra-conf: |
+            accept-flake-config = true
+            fallback = true
+            access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
+
+      - uses: DeterminateSystems/magic-nix-cache-action@v14
+        continue-on-error: true
+
+      - name: Connect to Tailscale
+        uses: tailscale/github-action@v4
```
0xgleb: should fix: these new workflows reference actions by mutable tags (@v4, @v22, @v14) while the deleted workflows pinned full commit SHAs. This step hands TS_AUTHKEY to tailscale/github-action@v4 and the nix-installer runs as root, so a moved tag on either could exfiltrate the tailnet key or backdoor the build with production credentials. Pin these to full SHAs with the version as a trailing comment, matching rainix.yaml.

infra/default.nix
```diff
@@ -0,0 +1,398 @@
...
+          }
+
+          _restart_service() {
+            if [ "$stay_stopped" = false ]; then
+              echo "Ensuring st0x-issuance is restarted on ${env}..." >&2
+              ssh_remote "mkdir -p /run/st0x && touch /run/st0x/st0x-issuance.ready && systemctl start st0x-issuance" || true
+            fi
+            _cleanup_identity
+          }
+          trap '_restart_service' EXIT
```
0xgleb, line 308: should fix: this replaces the earlier trap _cleanup_identity EXIT with _restart_service, and both success paths then run trap - EXIT without calling _cleanup_identity. When --op is used, the 1Password-sourced temp private key is left on disk on the happy path. A single EXIT trap that runs both, or an explicit _cleanup_identity before trap - EXIT, fixes it.

keys.nix
```diff
@@ -0,0 +1,69 @@
...
+    host-prod = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+    # PLACEHOLDER: replace with real key after `nix run .#stagingBootstrap`.
+    host-staging = "ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
+  };
+
+  roles = with keys; {
+    # access to terraform state and encrypted vars (shared across environments)
+    infra = [
```
0xgleb, line 37: should fix: the ci key is in roles.infra, and infra/secrets.nix encrypts terraform.tfstate and terraform.tfvars (which holds the DO API token) to that role. But the deploy workflows resolve the host via MagicDNS and never run Terraform, so CI does not need this. If secrets.SSH_KEY leaks, an attacker gets the DO API token and can rewrite the firewall rules. Drop ci from roles.infra; add a scoped role later if a CI Terraform job ever needs it.

.github/workflows/rekey.yaml
```diff
@@ -0,0 +1,69 @@
...
+        run: |
+          nix run github:yaxitech/ragenix -- \
+            --rules ./secret/secrets.nix \
+            -i ~/.ssh/id_ed25519 \
+            -r
```
0xgleb: should fix: ragenix rekey has to decrypt every secret before re-encrypting, so the identity must already be a recipient. The service secrets are encrypted to [st0x-op, host-prod, host-staging] and ci is not among them, so if secrets.SSH_KEY is the ci key this fails every run. The liquidity rekey only touches infra secrets that ci can actually decrypt

.env.secrets.example
```diff
@@ -0,0 +1,29 @@
...
+#   5. Delete plaintext: rm /tmp/st0x-issuance.env
+#
+# The encrypted file is committed at secret/st0x-issuance.env.age.
+# The Fireblocks RSA private key is a separate secret: secret/fireblocks-secret-issuance.key.age
+
+RPC_URL=${RPC_URL}
+FIREBLOCKS_API_USER_ID=${FIREBLOCKS_API_USER_ID}
+ISSUER_API_KEY=${ISSUER_API_KEY}
+ALPACA_IP_RANGES=${ALPACA_IP_RANGES}
+INTERNAL_IP_RANGES=127.0.0.0/8,::1/128,172.16.0.0/12,${INTERNAL_IP_RANGES_EXTRA}
```
0xgleb: should fix: the old deploy ran envsubst with the INTERNAL_IP_RANGES_EXTRA secret, but the agenix path does no substitution, so this literal ${INTERNAL_IP_RANGES_EXTRA} ends up in the encrypted file and the whitelist parser rejects it at startup. Replace it with concrete CIDR examples and note that extra ranges get baked in before encryption.

rust.nix
```diff
@@ -0,0 +1,108 @@
...
+    buildInputs = [
+      pkgs.openssl
+      pkgs.sqlite
+    ]
+
+    ++ pkgs.lib.optionals pkgs.stdenv.hostPlatform.isDarwin [
+      pkgs.apple-sdk_15
+    ];
+
+    DATABASE_URL = "sqlite:///tmp/build_db.sqlite";
```
0xgleb: should fix: this sets DATABASE_URL to a live sqlite path with no SQLX_OFFLINE, and the buildDepsOnly derivation inherits it without a preBuild to create the DB, so any dependency with compile-time-checked queries (apalis-sqlite is the documented case) tries to connect to a missing file. AGENTS.md and the liquidity rust.nix both build offline against the committed .sqlx/ cache. Set SQLX_OFFLINE = "true" in depsArgs and drop the live DATABASE_URL plus the sqlx preBuild.

os.nix
```diff
@@ -0,0 +1,173 @@
...
+  programs.bash.interactiveShellInit = "set -o vi";
+
+  systemd.tmpfiles.rules = [
+    "d /mnt/data 0755 st0x st0x -"
+    "d /mnt/data/logs 0755 st0x st0x -"
+  ];
+
+  # The system bus implementation cannot be live-switched safely. Deploy this
+  # change with deploy-rs --boot and a reboot, not a normal switch.
+  services.dbus.implementation = "broker";
```
0xgleb, line 157: should fix: this broker switch cannot be applied by a live switch-to-configuration, but prodDeployAll/stagingDeployAll do not use --boot. On the first deploy from bootstrap (which does not set broker) the live switch can fail activation and roll back to the bootstrap generation with no running service. Same root cause as the deploy-prod note: either set broker in bootstrap.nix, or run the boot deploy plus reboot before the service profiles.

nix/upgradeable-services.nix
```diff
@@ -0,0 +1,109 @@
...
+        # "staging" exposes them for easier debugging on the staging host.
+        "ENVIRONMENT=${
+          if environment == "prod" then
```
0xgleb, line 38: should fix: the unit sets ENVIRONMENT here via Environment=, but also pulls EnvironmentFile= from the shared env file, and .env.secrets.example includes ENVIRONMENT=production. systemd applies EnvironmentFile after Environment, so if an operator fills the template the staging host silently runs as production and hides the Swagger surface staging exists to expose. Drop ENVIRONMENT from the template (and note it is injected by the unit), or move it to a drop-in that applies after EnvironmentFile.

.env.secrets.example
0xgleb: nit: this says cp example.secrets.env ... but the file is .env.secrets.example, so anyone following it literally hits 'No such file or directory'.

.github/workflows/rekey.yaml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Setup SSH key
+        run: |
+          mkdir -p ~/.ssh
+          chmod 700 ~/.ssh
+          echo "${{ secrets.SSH_KEY }}" > ~/.ssh/id_ed25519
+          chmod 600 ~/.ssh/id_ed25519
+
+      - name: Rekey agenix secrets
+        run: |
+          nix run github:yaxitech/ragenix -- \
```
0xgleb: nit: this runs nix run github:yaxitech/ragenix at floating HEAD instead of the flake-pinned ragenix used everywhere else, which can drift and produce incompatible rekeys. nix run .#rekey uses the locked version.

JuaniRios (changes requested):
Two extra items that didn't fit on a changed line: .github/workflows/rainix.yaml line 7 sets cancel-in-progress on refs/heads/master, but this repo's default branch is main, so the concurrency guard silently never matches (pre-existing, worth fixing while you're in this file). And README.md still points to deleted deployment files (Dockerfile, docker-compose, rollback.sh) that this PR removes, so update it in the same pass.

0xgleb (changes requested):

.github/workflows/deploy-prod.yaml
rouzwelt: this workflow steps that are under `broker_migration mode` workflow input do not apply to issuance repo.

nix/upgradeable-services.nix
rouzwelt: left as is for now

deploy.nix
rouzwelt: thats an intentional design as far as I understand it, noted by Juan in other comments too and is similar to liquidity behavior

JuaniRios:
* **#204** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/204?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#211](https://github.com/ST0x-Technology/st0x.issuance/pull/211) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/211?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#234](https://github.com/ST0x-Technology/st0x.issuance/pull/234) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/234?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/204?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=204).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

JuaniRios (approved):

0xgleb (approved):

PR #161, opened 2026-05-20
feat: enrich stuck redemption tx metadata

JuaniRios:
## Motivation

The stuck admin endpoint reported redemptions with short generated `red-...` aggregate IDs and sparse metadata, while mint failures exposed richer transaction context. That made failed redemption investigation harder and left newly detected redemptions with weaker collision resistance.

Tracks [RAI-631](https://linear.app/makeitrain/issue/RAI-631/expose-full-tx-hash-metadata-for-stuck-redemptions).

## Solution

- Change new `IssuerRedemptionRequestId` values to serialize as the full detected on-chain transaction hash.
- Keep legacy `red-...` redemption IDs parseable and serializable so existing aggregates and callbacks continue to work.
- Enrich `/admin/stuck` redemption entries from event history with tokenization request ID, underlying, quantity, detected transfer tx hash, and Fireblocks tx ID when available.
- Add a `tx_hash` field to stuck entries and populate it for both mint and redemption failures when known, including Fireblocks status lookups that expose a network tx hash.
- Update `SPEC.md` to document full-hash redemption IDs and legacy compatibility.

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

Tested with:

- `nix develop -c cargo check --workspace`
- `nix develop -c cargo test -q --lib`
- `nix develop -c cargo test -q --workspace`
- `nix develop -c cargo clippy --workspace --all-targets --all-features -- -D clippy::all -D warnings`
- `nix develop -c cargo fmt --all -- --check`
- `PATH="$HOME/.cargo/bin:$PATH" nix develop -c cargo nextest run --workspace --retries 2 --flaky-result fail --no-fail-fast --status-level retry --final-status-level flaky --failure-output final`

## Summary by CodeRabbit

* **New Features**
  * Redemption request IDs now use full transaction hash format (`0x...`) instead of shortened identifiers.
  * Admin recovery endpoint updated to accept and return full transaction hash format for redemption IDs.

* **Documentation**
  * Updated specification and API documentation to reflect new redemption identifier format.
  * Added backward compatibility notes for historical redemptions using legacy identifier format.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ST0x-Technology/st0x.issuance/pull/161?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #211, opened 2026-07-03
feat: crash-safe burn idempotency via pre-signed tx

rouzwelt:
## Motivation

- The alloy (local signer) burn path had no crash safety. `submit_burn` signed, sent, and waited for the receipt in one go, so if the process died after the tx landed on-chain but before `BurnFireblocksSubmitted` was persisted, startup recovery would sign and send a brand new burn: double-burn.
- Fireblocks got this for free through deterministic `externalTxId` dedup. Alloy had nothing equivalent, and alloy is the path we're keeping now that Fireblocks is on its way out.
- Same bug class as [RAI-374](https://linear.app/makeitrain/issue/RAI-374/idempotent-mintburn-recovery-split-submitconfirm-events-phase-2), which fixed the Fireblocks side.

## Solution

- New `IntendBurn` command / `BurnIntended` event. `prepare_burn_tx()` builds and signs the multicall (redeems + optional dust transfer) and we persist the raw signed bytes, tx hash, nonce, and dust **before** any broadcast.
- **State machine change:** `BurnTokens` is now only valid from `BurnIntended`, not from `Burning`. Every burn goes through `IntendBurn` first.
- Local `submit_burn` now only broadcasts the persisted bytes via `eth_sendRawTransaction` and errors with `ExpectedPreparedTx` if nothing was prepared. A prepared tx is immutable: we never sign a replacement while its outcome is unknown. Fireblocks `submit_burn` ignores the prepared tx and behaves as before.
- Recovery from `BurnIntended { sendable_tx: Some(_) }` re-broadcasts the exact same bytes, then confirms against the stored hash. If the broadcast errors but `eth_getTransactionByHash` shows the node already has it, we treat it as sent.
- An ambiguous broadcast failure keeps the redemption in `BurnIntended` with its receipt reservation held, instead of releasing and re-signing. Operator resolves.
- `BurnIntended { sendable_tx: None }` is any backend whose `prepare_burn_tx` returns `None` (today, Fireblocks, via the trait default). It falls through to the existing submit + confirm flow.
- `RealBlockchainService` is no longer generic over `P`, it's pinned to one concrete filler stack so the tx we sign is the tx we hashed. Same stack now in `config.rs`, `tests/harness::create_provider()`, and the service unit tests, which is most of the test churn.
- New wallet nonce lock (`VaultService::lock_wallet()` → `WalletNonceGuard`) serializing nonce assignment through broadcast, plus a stale-state check after acquiring it so a queued burn doesn't re-execute. `submit_mint` takes the same lock around its send, so mints and burns can't race on the nonce.
- `confirm_burn` drops the in-process burn cache and always reads the receipt from chain via `PendingTransactionBuilder` (120s timeout). `MultiBurnParams` gains `detected_block_number`.

## Checks

## Anything else

New unit tests in `burn_manager.rs` cover recovery re-broadcasting the persisted tx, an ambiguous broadcast keeping `BurnIntended` + its reservation, `sendable_tx: None` recovery running the full flow, and `prepare_tx` failure releasing the reservation and recording `BurningFailed`. `redemption/mod.rs` covers `IntendBurn` from `Burning`, from a wrong state, and the `BurnIntended` apply. Mock gets `with_prepared_tx` and `new_prepare_tx_failure`. No new e2e: the crash window itself isn't exercised end to end, the existing e2e suites just moved to `create_provider()` for the new filler stack.

`BurnIntended` is a new permanent event and there's no migration. Redemptions already sitting in `Burning`/`BurnSubmitted` still recover through their old paths.

I added two `#[allow(clippy::too_many_lines)]` (`execute_burn_and_record_result` and `Redemption::apply_event`). Can split those functions instead if you'd rather not carry the allows.

## Summary by CodeRabbit

* **New Features**
  * Added a new burn “intention” phase that records the signed burn details before any broadcast.
  * Added support for deterministic, typed transaction identifiers to improve tracking across events and reporting.
* **Bug Fixes**
  * Improved burn execution/recovery to avoid duplicate processing and to re-use persisted signed burn bytes when outcomes are unclear.
  * Standardized transaction-id handling in admin stuck history and redemption history summaries.
* **Documentation**
  * Updated the redemption lifecycle and admin terminalization spec to include the new burn-intention step.
* **Tests**
  * Updated and extended coverage for the new burn-intention state and typed transaction identifiers.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -377,6 +377,13 @@ on-chain transfer through calling Alpaca to burning tokens.
   balance fails to reserve and never submits an unbacked burn. On confirmation
   the reservation is settled (mirror balance reduced); on a definitive
   terminal/reverted failure it is released.
+- `BurnIntended` - For a locally signed burn, persists the exact signed raw
+  transaction and its hash before any broadcast attempt. A prepared
+  transaction is immutable: live execution and recovery may only broadcast
+  those persisted bytes, never sign a replacement while its outcome is
+  unknown. Recovery re-broadcasts the same bytes before confirming the stored
+  hash. If the broadcast outcome remains ambiguous, the redemption stays in
+  `BurnIntended` with its receipt reservation held for operator resolution.
 - `TokensBurned` - On-chain burn succeeded, redemption complete (terminal
   success). Payload contains `burns: Vec<BurnRecord>` where each `BurnRecord`
   has `receipt_id` and `shares_burned`, supporting multi-receipt burns when a
@@ -409,6 +416,7 @@ on-chain transfer through calling Alpaca to burning tokens.
 | `RecordAlpacaCall`      | `AlpacaCalled`            | Alpaca API called                             |
 | `RecordAlpacaFailure`   | `AlpacaCallFailed`        | Terminal failure                              |
 | `ConfirmAlpacaComplete` | `AlpacaJournalCompleted`  | Journal complete                              |
+| `IntendBurn`            | `BurnIntended`            | Persist exact signed tx before broadcasting   |
 | `BurnTokens`            | `BurnFireblocksSubmitted` | Submits to signing backend                    |
 | `ConfirmBurn`           | `TokensBurned`            | Confirms burn, terminal success               |
 | `RecordBurnFailure`     | `BurningFailed`           | Records failure with optional tx metadata     |
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -252,7 +252,9 @@ async fn load_reprocess_context(
                 ..
             } => {
                 burning_failed = Some(BurningFailedData {
-                    fireblocks_tx_id: fireblocks_tx_id.clone(),
+                    fireblocks_tx_id: fireblocks_tx_id
+                        .as_ref()
+                        .map(ToString::to_string),
                     planned_burns: planned_burns.clone(),
                 });
             }
@@ -644,7 +646,9 @@ async fn inspect_prior_fireblocks_burn(
                     issuer_request_id,
                     RedemptionCommand::RecordExistingBurn {
                         issuer_request_id: issuer_request_id.clone(),
-                        fireblocks_tx_id: fb_tx_id.clone(),
+                        fireblocks_tx_id: crate::vault::TxId::Legacy(
+                            fb_tx_id.clone(),
+                        ),
                         tx_hash,
                         planned_burns: bf_data.planned_burns.clone(),
                         block_number,
@@ -1612,7 +1616,7 @@ fn redemption_history_summary_from_events(
                 fireblocks_tx_id: Some(fireblocks_tx_id),
                 ..
             } => {
-                summary.fireblocks_tx_id = Some(fireblocks_tx_id);
+                summary.fireblocks_tx_id = Some(fireblocks_tx_id.to_string());
             }
             _ => {}
         }
@@ -1934,7 +1938,9 @@ mod tests {
     use crate::test_utils::logs_contain_at;
     use crate::tokenized_asset::{TokenSymbol, UnderlyingSymbol};
     use crate::vault::mock::MockVaultService;
-    use crate::vault::{FireblocksTxStatus, MultiBurnEntry, VaultService};
+    use crate::vault::{
+        FireblocksTxStatus, MultiBurnEntry, SendableTxWithHash, VaultService,
+    };
 
     use super::{AggregateKind, StuckAggregate};
 
@@ -2356,6 +2362,30 @@ mod tests {
         store
             .send(
                 &metadata.issuer_request_id,
+                RedemptionCommand::IntendBurn {
+                    issuer_request_id: metadata.issuer_request_id.clone(),
+                    vault: address!(
+                        "0xcccccccccccccccccccccccccccccccccccccccc"
+                    ),
+                    burns: vec![MultiBurnEntry {
+                        receipt_id: U256::from(99),
+                        burn_shares: U256::from(100),
+                        receipt_info: None,
+                        receipt_info_bytes: None,
+                    }],
+                    dust_shares: U256::ZERO,
+                    owner: Address::ZERO,
+                    external_tx_id: Some(BurnExternalTxId::base(
+                        &metadata.detected_tx_hash,
+                    )),
+                },
+            )
+            .await
+            .expect("IntendBurn failed");
+
+        store
+            .send(
+                &metadata.issuer_request_id,
                 RedemptionCommand::BurnTokens {
                     issuer_request_id: metadata.issuer_request_id.clone(),
                     vault: address!(
@@ -2383,7 +2413,9 @@ mod tests {
                 RedemptionCommand::RecordBurnFailure {
                     issuer_request_id: metadata.issuer_request_id.clone(),
                     error: "burn terminally failed".to_string(),
-                    fireblocks_tx_id: Some("mock-fb-burn".to_string()),
+                    fireblocks_tx_id: Some(crate::vault::TxId::Legacy(
+                        "mock-fb-burn".to_string(),
+                    )),
                     planned_burns: vec![],
                 },
             )
@@ -3483,14 +3515,20 @@ mod tests {
         async fn submit_burn(
             &self,
             _params: crate::vault::MultiBurnParams,
-        ) -> Result<crate::vault::SubmittedTx, crate::vault::VaultError>
-        {
+            _prepared_tx: Option<SendableTxWithHash>,
+        ) -> Result<
+            crate::vault::SubmittedTx<
+                crate::redemption::BurnExternalTxId,
+                crate::vault::TxId,
+            >,
+            crate::vault::VaultError,
+        > {
             unimplemented!("not used in enrichment tests")
         }
 
         async fn confirm_burn(
             &self,
-            _fireblocks_tx_id: &str,
+            _fireblocks_tx_id: &crate::vault::TxId,
             _dust_shares: alloy::primitives::U256,
         ) -> Result<crate::vault::MultiBurnResult, crate::vault::VaultError>
         {
@@ -4171,7 +4209,9 @@ mod tests {
                 external_tx_id: BurnExternalTxId::from_string(format!(
                     "burn-{tx_hash}"
                 )),
-                fireblocks_tx_id: "fb-old-attempt".to_string(),
+                fireblocks_tx_id: crate::vault::TxId::Legacy(
+                    "fb-old-attempt".to_string(),
+                ),
                 planned_burns: vec![],
                 submitted_at: now,
             },
@@ -4179,7 +4219,9 @@ mod tests {
                 issuer_request_id: issuer.clone(),
                 error: "fireblocks failed".to_string(),
                 failed_at: now,
-                fireblocks_tx_id: Some("fb-old-attempt".to_string()),
+                fireblocks_tx_id: Some(crate::vault::TxId::Legacy(
+                    "fb-old-attempt".to_string(),
+                )),
                 planned_burns: vec![],
             },
             RedemptionEvent::RedemptionFailed {
--- a/src/redemption/burn_manager.rs
+++ b/src/redemption/burn_manager.rs
@@ -19,12 +19,13 @@ use crate::receipt_inventory::{
     BurnPlan, BurnTrackingError, ReceiptRegistrationError, ReceiptService,
     Shares,
 };
+use crate::redemption::{BurnRecord, RedemptionMetadata};
 use crate::tokenized_asset::UnderlyingSymbol;
 use crate::tokenized_asset::view::{
     TokenizedAssetViewError, find_vault_by_underlying,
 };
 use crate::vault::{
-    BurnVerification, FireblocksTxStatus, MultiBurnEntry, VaultError,
+    BurnVerification, FireblocksTxStatus, MultiBurnEntry, TxId, VaultError,
     VaultService,
 };
 
@@ -161,7 +162,8 @@ impl BurnManager {
         self.recover_single_burning(issuer_request_id).await
     }
 
-    /// Admin-terminalizes a redemption stuck in `Burning`/`BurnSubmitted` whose
+    /// Admin-terminalizes a redemption stuck in
+    /// `Burning`/`BurnIntended`/`BurnSubmitted` whose
     /// burn already landed on-chain but was never recorded (e.g. a crash
     /// between the burn and `TokensBurned`).
     ///
@@ -191,6 +193,7 @@ impl BurnManager {
 
         let underlying = match &redemption {
             Redemption::Burning { metadata, .. }
+            | Redemption::BurnIntended { metadata, .. }
             | Redemption::BurnSubmitted { metadata, .. } => {
                 metadata.underlying.clone()
             }
@@ -533,7 +536,10 @@ impl BurnManager {
 
         match self
             .vault_service
-            .confirm_burn(fireblocks_tx_id, dust_shares)
+            .confirm_burn(
+                &TxId::Legacy(fireblocks_tx_id.to_string()),
+                dust_shares,
+            )
             .await
         {
             Ok(result) => {
@@ -561,7 +567,9 @@ impl BurnManager {
                         issuer_request_id,
                         RedemptionCommand::RecordExistingBurn {
                             issuer_request_id: issuer_request_id.clone(),
-                            fireblocks_tx_id: fireblocks_tx_id.to_string(),
+                            fireblocks_tx_id: TxId::Legacy(
+                                fireblocks_tx_id.to_string(),
+                            ),
                             tx_hash: result.tx_hash,
                             planned_burns: actual_burns,
                             block_number: result.block_number,
@@ -607,6 +615,107 @@ impl BurnManager {
         }
     }
 
+    async fn recover_single_burning_shared_inner(
+        &self,
+        issuer_request_id: &IssuerRedemptionRequestId,
+        metadata: &RedemptionMetadata,
+        tx_id: &TxId,
+        dust_shares: U256,
+        planned_burns: &[BurnRecord],
+        recovery_state: BurnRecoveryState,
+    ) -> Result<RecoveryOutcome, BurnManagerError> {
+        let vault =
+            find_vault_by_underlying(&self.view_pool, &metadata.underlying)
+                .await?
+                .ok_or_else(|| BurnManagerError::AssetNotFound {
+                    underlying: metadata.underlying.clone(),
+                })?;
+
+        if recovery_state == BurnRecoveryState::Submitted {
+            info!(target: "redemption", issuer_request_id = %issuer_request_id,
+                fireblocks_tx_id = %tx_id,
+                "Recovering BurnSubmitted redemption - confirming existing transaction"
+            );
+        } else {
+            info!(target: "redemption", issuer_request_id = %issuer_request_id,
+                fireblocks_tx_id = %tx_id,
+                "Recovering BurnIntended redemption - checking existing transaction"
+            );
+        }
+
+        let confirm_result = self
+            .store
+            .send(
+                issuer_request_id,
+                RedemptionCommand::ConfirmBurn {
+                    issuer_request_id: issuer_request_id.clone(),
+                    fireblocks_tx_id: tx_id.clone(),
+                    dust_shares,
+                },
+            )
+            .await;
+
+        match confirm_result {
+            Ok(()) => {
+                info!(target: "redemption", issuer_request_id = %issuer_request_id,
+                    "Burn confirmed successfully during recovery"
+                );
+
+                self.settle_reserved_burn(vault, issuer_request_id).await;
+
+                Ok(RecoveryOutcome::ExistingBurnRecorded)
+            }
+            Err(AggregateError::UserError(LifecycleError::Apply(
+                RedemptionError::Vault {
+                    message: err,
+                    release_reservation,
+                    ..
+                },
+            ))) => {
+                if recovery_state == BurnRecoveryState::Submitted {
+                    warn!(target: "redemption", issuer_request_id = %issuer_request_id,
+                        error = %err,
+                        "Burn confirmation failed during recovery"
+                    );
+                } else {
+                    warn!(target: "redemption", issuer_request_id = %issuer_request_id,
+                        error = %err,
+                        "BurnIntended confirmation failed during recovery"
+                    );
+                }
+
+                if release_reservation {
+                    // Release before terminalizing the redemption. A crash
+                    // after this idempotent release leaves the aggregate in
+                    // its recoverable state, so the same transaction is
+                    // checked again. Recording failure first could strand a
+                    // reservation because burning-state recovery would no
+                    // longer revisit the aggregate.
+                    self.release_reserved_burn(vault, issuer_request_id).await;
+                }
+
+                self.store
+                    .send(
+                        issuer_request_id,
+                        RedemptionCommand::RecordBurnFailure {
+                            issuer_request_id: issuer_request_id.clone(),
+                            error: err.clone(),
+                            fireblocks_tx_id: Some(tx_id.clone()),
+                            planned_burns: planned_burns.to_vec(),
+                        },
+                    )
+                    .await?;
+
+                Err(BurnManagerError::Redemption(RedemptionError::Vault {
+                    message: err,
+                    release_reservation,
+                    fireblocks_tx_id: None,
+                }))
+            }
+            Err(err) => Err(err.into()),
+        }
+    }
+
     async fn recover_single_burning(
         &self,
         issuer_request_id: &IssuerRedemptionRequestId,
@@ -628,7 +737,24 @@ impl BurnManager {
                 ..
             } => {
                 let dust_shares = dust_quantity.to_u256_with_18_decimals()?;
+                self.recover_single_burning_shared_inner(
+                    issuer_request_id,
+                    metadata,
+                    fireblocks_tx_id,
+                    dust_shares,
+                    planned_burns,
+                    BurnRecoveryState::Submitted,
+                )
+                .await
+            }
 
+            Redemption::BurnIntended {
+                metadata,
+                planned_burns,
+                sendable_tx: Some(sendable_tx),
+                external_tx_id,
+                ..
+            } => {
                 let vault = find_vault_by_underlying(
                     &self.view_pool,
                     &metadata.underlying,
@@ -639,88 +765,76 @@ impl BurnManager {
                         underlying: metadata.underlying.clone(),
                     }
                 })?;
+                let dust_shares = sendable_tx.dust_shares;
+                let tx_id = TxId::Hash(sendable_tx.hash);
 
-                info!(target: "redemption", issuer_request_id = %issuer_request_id,
-                    fireblocks_tx_id = %fireblocks_tx_id,
-                    "Recovering BurnSubmitted redemption - confirming existing transaction"
+                info!(target: "redemption",
+                    issuer_request_id = %issuer_request_id,
+                    tx_hash = %sendable_tx.hash,
+                    "Recovering BurnIntended redemption - re-broadcasting persisted transaction"
                 );
 
-                let confirm_result = self
+                let burns = planned_burns
+                    .iter()
+                    .map(|burn| MultiBurnEntry {
+                        receipt_id: burn.receipt_id,
+                        burn_shares: burn.shares_burned,
+                        receipt_info: None,
+                        receipt_info_bytes: None,
+                    })
+                    .collect();
+                let wallet_guard = self.vault_service.lock_wallet().await;
+                let submit_result = self
                     .store
                     .send(
                         issuer_request_id,
-                        RedemptionCommand::ConfirmBurn {
+                        RedemptionCommand::BurnTokens {
                             issuer_request_id: issuer_request_id.clone(),
-                            fireblocks_tx_id: fireblocks_tx_id.clone(),
+                            vault,
+                            burns,
                             dust_shares,
+                            owner: self.bot_wallet,
+                            external_tx_id: external_tx_id.clone(),
                         },
                     )
                     .await;
+                drop(wallet_guard);
 
-                match confirm_result {
-                    Ok(()) => {
-                        info!(target: "redemption", issuer_request_id = %issuer_request_id,
-                            "Burn confirmed successfully during recovery"
-                        );
-
-                        self.settle_reserved_burn(vault, issuer_request_id)
-                            .await;
-
-                        Ok(RecoveryOutcome::ExistingBurnRecorded)
-                    }
-                    Err(AggregateError::UserError(LifecycleError::Apply(
-                        RedemptionError::Vault {
-                            message: err,
-                            release_reservation,
-                            ..
-                        },
-                    ))) => {
-                        warn!(target: "redemption", issuer_request_id = %issuer_request_id,
-                            error = %err,
-                            "Burn confirmation failed during recovery"
-                        );
-
-                        if release_reservation {
-                            self.release_reserved_burn(
-                                vault,
-                                issuer_request_id,
-                            )
-                            .await;
-                        }
-
-                        self.store
-                            .send(
-                                issuer_request_id,
-                                RedemptionCommand::RecordBurnFailure {
-                                    issuer_request_id: issuer_request_id
-                                        .clone(),
-                                    error: err.clone(),
-                                    fireblocks_tx_id: Some(
-                                        fireblocks_tx_id.clone(),
-                                    ),
-                                    planned_burns: planned_burns.clone(),
-                                },
-                            )
-                            .await?;
-
-                        Err(BurnManagerError::Redemption(
-                            RedemptionError::Vault {
-                                message: err,
-                                release_reservation,
-                                fireblocks_tx_id: None,
-                            },
-                        ))
-                    }
-                    Err(err) => Err(err.into()),
+                if let Err(error) = submit_result {
+                    warn!(target: "redemption",
+                        issuer_request_id = %issuer_request_id,
+                        tx_hash = %sendable_tx.hash,
+                        error = %error,
+                        "Persisted burn transaction re-broadcast failed; keeping BurnIntended for recovery"
+                    );
+                    return Err(error.into());
                 }
+
+                self.recover_single_burning_shared_inner(
+                    issuer_request_id,
+                    metadata,
+                    &tx_id,
+                    dust_shares,
+                    planned_burns,
+                    BurnRecoveryState::Intended,
+                )
+                .await
             }
 
-            // Still in Burning state — needs full submit + confirm flow
+            // Still in Burning/BurnIntended(with no prepared tx - mean fireblocks path) state — needs full submit + confirm flow
             Redemption::Burning {
                 metadata,
                 alpaca_quantity,
                 dust_quantity,
                 external_tx_id,
+                ..
+            }
+            | Redemption::BurnIntended {
+                metadata,
+                alpaca_quantity,
+                dust_quantity,
+                external_tx_id,
+                sendable_tx: None,
                 ..
             } => {
                 let vault = find_vault_by_underlying(
@@ -825,13 +939,21 @@ impl BurnManager {
         issuer_request_id: &IssuerRedemptionRequestId,
         aggregate: &Redemption,
     ) -> Result<(), BurnManagerError> {
-        let Redemption::Burning {
+        let (Redemption::Burning {
             metadata,
             alpaca_quantity,
             dust_quantity,
             external_tx_id,
             ..
-        } = aggregate
+        }
+        | Redemption::BurnIntended {
+            metadata,
+            alpaca_quantity,
+            dust_quantity,
+            external_tx_id,
+            sendable_tx: None,
+            ..
+        }) = aggregate
         else {
             return Err(BurnManagerError::InvalidAggregateState {
                 current_state: aggregate_state_name(aggregate).to_string(),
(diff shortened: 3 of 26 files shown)


rouzwelt:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/211?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#225** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/225?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#226](https://github.com/ST0x-Technology/st0x.issuance/pull/226) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/226?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#227](https://github.com/ST0x-Technology/st0x.issuance/pull/227) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/227?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#224** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#223** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/223?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#123** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/123?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#118** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/118?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#230** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/230?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#229** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/229?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#228** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#221** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/221?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#211** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/211?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/211?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#204** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/204?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#234](https://github.com/ST0x-Technology/st0x.issuance/pull/234) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/234?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=211).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/vault/service.rs
```diff
@@ -410,6 +365,62 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
+                });
+
+            return Ok(Some(SubmittedTx {
+                external_tx_id,
+                fireblocks_tx_id: tx_hash.to_string(),
+            }));
+        }
+
+        Ok(None)
+    }
```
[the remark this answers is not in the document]
rouzwelt: Looks fine because burns are executed as an atomic multicall, a `Withdraw` for any receipt ID in the batch implies the whole batch was burned in that tx.
but will wait on @0xgleb and @JuaniRios to check and comment too

src/vault/service.rs
```diff
@@ -410,6 +365,71 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
+            for log in &logs {
+                let Some(tx_hash) = log.transaction_hash else { continue };
+                let Ok(decoded) =
+                    log.log_decode::<OffchainAssetReceiptVault::Withdraw>()
+                else {
+                    continue;
+                };
+
+                let event_data = decoded.data();
+                if event_data.owner != params.owner
```
JuaniRios: critical: this treats any Withdraw from the bot wallet touching any one of the batch's receipt IDs as proof the whole batch already burned. But `owner` is always the bot wallet so it does not discriminate between redemptions, and receipt IDs are reusable partial-burn buckets, so a different concurrent redemption that burns an overlapping receipt after `detected_block_number` will false-match here. On a false match we persist BurnFireblocksSubmitted, confirm the wrong tx, and complete this redemption without its shares actually being burned, which breaks 1:1 backing. Match the full candidate tx before returning it: vault, owner, receiver, and the exact multiset of planned (receipt_id, shares) plus dust.

src/vault/service.rs
```diff
@@ -338,15 +299,9 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
-        }
-
-        // Cache empty (process restarted) — re-fetch receipt from chain using tx hash
+        // Fetch receipt from chain using tx hash
         debug!(target: "vault", tx_hash = %fireblocks_tx_id,
-            "Burn cache empty, recovering from chain"
+            "Getting burn tx data from chain"
         );
 
         let tx_hash: B256 =
```
JuaniRios: should fix: submit_burn now returns right after broadcast without waiting for a receipt, but confirm_burn does a single get_transaction_receipt and maps the not-yet-mined case to InvalidReceipt. The burn manager sends ConfirmBurn immediately after submit, so on any real chain (non-instamine) a healthy pending burn gets recorded as BurningFailed. Either keep the alloy submit_burn synchronous until a receipt exists, or add an explicit pending outcome that leaves the aggregate in BurnSubmitted with the tx hash preserved for a later retry. The module doc claims confirm_burn polls, but there is no poll loop here.

src/vault/service.rs
```diff
@@ -410,6 +365,71 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
+    /// receipt ID in the batch implies the whole batch was burned in that tx.
+    /// Prevents double-burning when the process crashes after the tx confirms
+    /// on-chain but before `BurnSubmitted` is persisted to the CQRS store.
+    async fn find_existing_burn(
+        &self,
+        params: &super::MultiBurnParams,
+    ) -> Result<Option<SubmittedTx>, VaultError> {
+        let current_block_number = self.provider.get_block_number().await?;
+        let mut start_block_number = params.detected_block_number;
+
```
JuaniRios: should fix: this scan only sees mined Withdraw logs up to the current block, so it cannot see a tx still in the mempool or one that mines just after the scan. If the crash happened after broadcast but before persisting BurnFireblocksSubmitted, find_existing_burn returns None and recovery submits a second burn while the original can still land. That is exactly the double-burn window the PR claims to close, so it is only partially closed. Closing it fully needs durable tx identity persisted at/before broadcast (signed tx hash or nonce) that recovery can poll, not just a mined-log scan.

src/redemption/burn_manager.rs
```diff
@@ -506,7 +506,8 @@ impl BurnManager {
             });
         };
 
-        self.handle_burning_started(issuer_request_id, &aggregate).await?;
+        self.handle_burning_started(issuer_request_id, &aggregate, false)
```
JuaniRios: should fix: when find_existing_burn's scan itself fails the burn goes to BurnFailed, and this retry path calls handle_burning_started with try_find_tx=false. So the retry that runs right after a failed idempotency check skips the check entirely and can resubmit a burn that already landed. The retry after a burn we could not confirm is exactly when try_find_tx should be true.

src/redemption/mod.rs
```diff
@@ -409,23 +410,58 @@ impl Redemption {
...
+            issuer_request_id: issuer_request_id.clone(),
+            detected_tx_hash: metadata.detected_tx_hash,
+            external_tx_id: input.external_tx_id,
+            detected_block_number: metadata.block_number,
+        };
+
+        // if try_find_tx is true, try to find an existing burn first to avoid burn duplication,
+        // this is the case for startup recovery process when a tx was submitted but process crash
+        // before persisting the submitted tx sate into store
+        if try_find_tx {
```
JuaniRios: should fix: treating find_existing_burn == Ok(None) as proof it is safe to submit assumes the scan can see every prior attempt, but it only sees mined logs. In the crash-after-broadcast-before-persist window the original tx is still pending, so None here means resubmit, not safe. This is the same root cause as the service.rs scan comment: the decision needs durable pre-broadcast tx identity, not absence of a mined log.

src/redemption/burn_manager.rs
```diff
@@ -904,6 +911,7 @@ impl BurnManager {
             vault,
```
JuaniRios, line 1026: should fix: recovery calls plan_burn and reserve_with_conflict_retry before find_existing_burn runs. If the burn already landed, replanning against current receipt balances can produce a different plan (or contend on reservations) than what was actually burned, so even a successful find match may be reconciled against a mismatched plan. Scan for the existing burn first, and only plan/reserve when nothing is found.

src/vault/service.rs
```diff
@@ -410,6 +365,71 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
+        // check logs in batch 5 blocks at a time
+        while start_block_number <= current_block_number {
+            let filter = Filter::new()
+                .address(params.vault)
+                .from_block(start_block_number)
+                .to_block((current_block_number).min(start_block_number + 5))
```
JuaniRios: minor: the scan walks detected_block_number to head in fixed 5-block windows, one sequential get_logs per window. A redemption that sat in Burning for a while turns startup recovery into a long serial burst of RPC calls, and given past dRPC exhaustion that is a real cost. Use a much larger window, bound the upper end, or look the tx up directly once identity is persisted.

src/vault/service.rs
```diff
@@ -996,4 +1023,355 @@ mod tests {
...
+        asserter.push_success(&0u64);
+        asserter.push_success(&Vec::<alloy::rpc::types::Log>::new());
+
+        let service = create_service_with_asserter(asserter);
+        let result = service
+            .find_existing_burn(&MultiBurnParams {
+                vault: test_vault_address(),
+                burns: vec![MultiBurnEntry {
+                    receipt_id: U256::from(1),
+                    burn_shares: U256::from(100),
```
JuaniRios: should fix: every find_existing_burn test sets detected_block_number = current, so the while loop runs a single iteration. The multi-window pagination (5-block stride and boundary re-scan) is never exercised, which is where off-by-one and missed-log bugs would hide. Add a test where the matching Withdraw is several windows past the detected block.

src/redemption/burn_manager.rs
```diff
@@ -777,8 +778,13 @@ impl BurnManager {
                     "Recovering Burning redemption - resuming burn"
                 );
 
-                self.handle_burning_started(issuer_request_id, &aggregate)
-                    .await?;
+                // set try_find_tx to true so it first tries to find the tx before resubmitting
+                self.handle_burning_started(
```
JuaniRios: should fix: the point of this PR is that recovery now calls handle_burning_started with try_find_tx=true, but no test drives a Burning recovery through BurnManager with an existing on-chain burn present. The one behavior that prevents the double-burn is untested at this layer. Add a recovery test whose mock returns an existing burn and assert submit_burn is not called.

src/redemption/cmd.rs
```diff
@@ -52,6 +52,15 @@ pub(crate) enum RedemptionCommand {
...
         external_tx_id: Option<BurnExternalTxId>,
+        /// Determines if should try searching to recover a transaction that
+        /// was not recorded into store as a submitted tx, this can happen
+        /// when a transaction is submitted onchain but process have crashed
+        /// before persisting the submitted tx state into the store.
+        /// Generally this is used for alloy path, since Fireblocks path handles
+        /// this through unique `external_tx_id` remote lookup.
+        ///
+        /// VaultServices handle this through `find_existing_burn()`.
+        try_find_tx: bool,
```
JuaniRios: minor: try_find_tx is a bare bool threaded unchanged through the command, the aggregate handler, and two BurnManager methods. A bool at a call site reads as a mystery flag. A small two-variant enum (e.g. Fresh / RecoverExisting) would make intent legible at each call.

src/redemption/cmd.rs
```diff
@@ -52,6 +52,15 @@ pub(crate) enum RedemptionCommand {
         /// Fireblocks burn terminally failed.
         #[serde(default)]
         external_tx_id: Option<BurnExternalTxId>,
+        /// Determines if should try searching to recover a transaction that
```
JuaniRios: minor: this changes the BurnTokens command contract with the new try_find_tx field, and AGENTS.md asks for SPEC.md to be updated when commands change. Worth a line in SPEC so the command surface stays accurate.

src/vault/service.rs
```diff
@@ -874,6 +894,7 @@ mod tests {
                 issuer_request_id: test_issuer_redemption_id(),
                 detected_tx_hash,
                 external_tx_id: None,
+                detected_block_number: 0,
             })
             .await;
```
JuaniRios, line 942: nit: the negated `.all(|burn| id != receipt_id)` reads awkwardly. `!params.burns.iter().any(|burn| event_data.id == burn.receipt_id)` says 'none of the planned receipts match' more directly.

src/redemption/burn_manager.rs
```diff
@@ -824,6 +830,7 @@ impl BurnManager {
         &self,
         issuer_request_id: &IssuerRedemptionRequestId,
         aggregate: &Redemption,
+        try_find_tx: bool,
```
JuaniRios: nit: the # Arguments doc for handle_burning_started still lists only issuer_request_id and aggregate, but there is now a third try_find_tx param that changes control flow. Add it to the doc.

src/vault/service.rs
```diff
@@ -23,22 +25,31 @@ use crate::redemption::BurnExternalTxId;
...
-/// `submit_mint`/`submit_burn` execute the full operation and cache the result for
-/// `confirm_mint`/`confirm_burn` to return. Results are keyed by tx hash to prevent
+/// `submit_mint` execute the full operation and cache the result for
+/// `confirm_mint` to return. Results are keyed by tx hash to prevent
 /// concurrent operations from overwriting each other.
+/// `submit_burn` and `confirm_burn` on the other hand do NOT follow this pattern,
+/// due to minimizing the chance of double burn as a result of a crash. We submit
+/// the tx and do not poll for receipt, that means the tx_hash is returned right
+/// away, state then records a submitted tx and polling for receipt happens in `confirm_burn`.
+/// ``
```
JuaniRios: nit: stray empty ``` line left in the doc comment here.

src/redemption/mod.rs
```diff
@@ -3347,4 +3387,169 @@ mod tests {
...
+        let receipt_id = uint!(42_U256);
+        let burn_shares = uint!(100_000000000000000000_U256);
+        let vault = address!("0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa");
+        let owner = address!("0x1111111111111111111111111111111111111111");
+        let user_wallet =
+            address!("0x9876543210fedcba9876543210fedcba98765432");
+        let found_external_tx_id = "burn-0xabcd-retry-1".to_string();
+        let found_fireblocks_tx_id = "fb-found-tx-999".to_string();
+
+        let services: Arc<dyn VaultService> = Arc::new(
```
JuaniRios: minor: this test infers submit_burn was skipped only because the found tx's canned IDs differ from what submit_burn would return. A mock that asserts if submit_burn is called on the try_find_tx=true path would test the skip directly.

JuaniRios (changes requested):

src/redemption/cmd.rs
0xgleb: should fix: try_find_tx: bool threaded through BurnTokens, handle_burning_started, execute_burn_and_record_result, and handle_burn_tokens is boolean blindness, every call site reads as a bare true or false, and the one recovery site even needs a comment to explain its literal. A two-variant enum like BurnRecoveryMode { AlwaysSubmit, TryFindExisting } makes the recovery call site self-evident and the rest obvious without comments.

src/redemption/mod.rs
```diff
@@ -409,23 +410,58 @@ impl Redemption {
...
+                            issuer_request_id,
+                            external_tx_id: BurnExternalTxId::from_string(
+                                found_tx.external_tx_id,
+                            ),
+                            fireblocks_tx_id: found_tx.fireblocks_tx_id,
+                            planned_burns,
+                            submitted_at: Utc::now(),
+                        },
+                    ]);
+                }
```
0xgleb: should fix: when the scan finds an existing burn this silently emits BurnFireblocksSubmitted and skips submit_burn. It's the one place the crash-recovery protection actually fires and there's no log at any level. Add an info! with the issuer_request_id and the discovered tx hash so an operator can distinguish a recovered burn from a fresh submit.

src/redemption/burn_manager.rs
```diff
@@ -777,8 +778,13 @@ impl BurnManager {
...
 
-                self.handle_burning_started(issuer_request_id, &aggregate)
-                    .await?;
+                // set try_find_tx to true so it first tries to find the tx before resubmitting
+                self.handle_burning_started(
+                    issuer_request_id,
+                    &aggregate,
+                    true,
+                )
+                .await?;
```
0xgleb: should fix: this recovery path (try_find_tx = true flowing through to find_existing_burn) is tested at the aggregate level but has no BurnManager-level test. A test seeding a Burning redemption with with_existing_burn(...) and asserting recover_burning_redemptions adopts it without calling submit_burn would cover the wiring. Also the comment above just restates the argument; with an enum instead of the bool it can go entirely.

src/redemption/mod.rs
```diff
@@ -409,23 +410,58 @@ impl Redemption {
...
+                Err(e) => {
+                    return Err(RedemptionError::Vault {
+                        release_reservation: false,
+                        fireblocks_tx_id: None,
+                        message: format!(
+                            "Failed to scan for existing burn due to: {e}"
+                        ),
+                    });
+                }
+            }
```
0xgleb: should fix: if the scan itself errors (any transient RPC failure) this records a burn failure, and the BurnFailed retry path calls handle_burning_started with try_find_tx = false, so the retry runs the unprotected resubmit in exactly the scenario the scan was added for. Either keep the redemption in Burning on scan errors so the next recovery re-runs the protected path, or propagate try_find_tx = true through the BurnFailed retry whenever a tx may already have been submitted.

src/vault/service.rs
```diff
@@ -23,22 +25,31 @@ use crate::redemption::BurnExternalTxId;
...
+/// **Crash-safe idempotency:**
+/// - **Mints**: handled at the application layer — the receipt backfiller runs at
+///   startup before recovery, scans the chain for `Deposit` events, and populates
+///   the receipt inventory by `issuer_request_id`. `recover_from_existing_receipt`
+///   finds the existing deposit and emits `ExistingMintRecovered`, preventing a
+///   double-mint. `submit_mint` also scans for an existing deposit before sending
+///   as defense-in-depth (covers periodic-backfill gaps during live operation).
+/// - **Burns**: through `find_existing_burn()` which scans for a matching `Withdraw`
+///   event by `receipt_id`. Because burns are executed as an atomic multicall,
+///   finding a `Withdraw` for any receipt in the batch implies the whole batch was
```
0xgleb: should fix: this doc now advertises crash-safe idempotency, but find_existing_burn only sees mined transactions. A crash between .send() and mining leaves a pending tx the Withdraw-event scan cannot find, and the old dev/test-only disclaimer was removed. Worth restating that caveat, since Fireblocks dedupes via externalTxId regardless of mining status and this path doesn't. While here: drop the stray empty backtick pair a couple of lines up and fix the "submit_mint execute" grammar.

src/vault/service.rs
```diff
@@ -410,6 +365,71 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
+                        ToString::to_string,
+                    );
+
+                return Ok(Some(SubmittedTx {
+                    external_tx_id,
+                    fireblocks_tx_id: tx_hash.to_string(),
+                }));
+            }
+
+            start_block_number += 5;
```
0xgleb: minor: the query range is inclusive (to_block = start + 5) but the stride is += 5, so every boundary block is scanned by two consecutive batches. Harmless for correctness but it re-fetches a block per batch and the "batch 5 blocks" comment doesn't match the actual 6-block window. Either advance by 6 or cap to_block at start + 4.

src/vault/service.rs
```diff
@@ -410,6 +365,71 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
+                        .all(|burn| event_data.id != burn.receipt_id)
+                {
+                    continue;
+                }
+
+                let external_tx_id =
+                    params.external_tx_id.as_ref().map_or_else(
+                        || {
+                            BurnExternalTxId::base(&params.detected_tx_hash)
+                                .into_string()
```
0xgleb: minor: this builds a BurnExternalTxId and immediately .into_string() to fit SubmittedTx, and the caller re-parses it with from_string. Round-tripping a domain type through String at an internal boundary drops the parse-don't-validate guarantee. Returning Option<(BurnExternalTxId, B256)> or a burn-specific struct keeps the type intact.

src/redemption/mod.rs
```diff
@@ -3347,4 +3387,169 @@ mod tests {
             "Startup must clear incompatible Redemption snapshots"
         );
     }
+
+    #[tokio::test]
+    async fn test_burn_tokens_uses_found_tx_when_try_find_tx_is_true() {
```
0xgleb: minor: neither new try_find_tx test uses #[traced_test] or asserts logs, and the Ok(None) fallthrough, the most common crash case where nothing was submitted and we must burn fresh, has no test at all. Worth a third test with no with_existing_burn asserting the IDs come from submit_burn, plus log assertions once the recovery info log exists.

src/vault/service.rs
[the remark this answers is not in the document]
0xgleb: manual burns might invalidate the multicall assumption

0xgleb (changes requested):

src/redemption/cmd.rs
rouzwelt: this has been addressed in last PR in the stack which is specifically for SPEC update

src/redemption/burn_manager.rs
rouzwelt: added `test_recover_burning_uses_existing_burn_without_resubmitting` test

src/redemption/event.rs
```diff
@@ -234,6 +235,11 @@ pub(crate) enum RedemptionEvent {
         external_tx_id: Option<BurnExternalTxId>,
         resumed_at: DateTime<Utc>,
     },
+    BurnIntended {
+        issuer_request_id: IssuerRedemptionRequestId,
+        sendable_tx: Option<SendableTxWithHash>,
+        planned_burns: Vec<BurnRecord>,
+    },
```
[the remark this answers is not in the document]
rouzwelt: in the sendable_tx there is a signed_at time, and for Fireblocks path there is no prepared tx before submission so th path doesn't need it anyways, The Option<> is going to be removed in next PR once Fireblocks is completely removed

src/redemption/burn_manager.rs
```diff
@@ -1131,6 +1265,22 @@ impl BurnManager {
...
+                    return Box::pin(self.execute_burn_and_record_result(
+                        issuer_request_id,
+                        vault,
+                        plan,
+                        external_tx_id,
+                        retry_count + 1,
+                    ))
+                    .await;
+                }
+
```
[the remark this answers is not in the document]
rouzwelt: any submission error means the tx did not actually land onchain, we can build a new tx and retry with max 5 times, the case u are describing is when the confirmation fails, ie the tx submitted, but we are not sure if it landed or is going to, if a crash happens after submission and before persisting the BurnSubmitted state on db, then upon recovery we check for the tx onchain (polling the tx for 30s) and if it was mined we record as success and if not we record as failure.

src/redemption/burn_manager.rs
```diff
@@ -1131,6 +1265,22 @@ impl BurnManager {
                     self.release_reserved_burn(vault, issuer_request_id).await;
                 }
 
+                // upon submission failure, if there was precomputed tx (alloy path)
+                // we rerun the process with max 5 retries, which ensures a new tx is
+                // built and persisted and if all 5 retries fail we proceed to record
+                // as failure.
+                // for fireblocks path we proceed normally to record failure
+                if sendable_tx.is_some() && retry_count < MAX_RETRIES {
```
0xgleb: critical: when a prepared raw tx submission fails ambiguously (Rpc, PendingTransaction -- anything except Reverted), this branch recurses into execute_burn_and_record_result, which calls prepare_burn_tx again and signs a brand-new tx with a fresh nonce. A send error does not mean the node does not have the tx: geth returns the error `already known` precisely when the pool *already contains* the submitted tx ([go-ethereum core/txpool/errors.go, `ErrAlreadyKnown`](https://github.com/ethereum/go-ethereum/blob/master/core/txpool/errors.go)), and a transport timeout / dropped connection can land *after* the node accepted the raw tx. Once any node has it, it is gossiped to peers and cannot be recalled ([devp2p eth spec, Transaction Exchange](https://github.com/ethereum/devp2p/blob/master/caps/eth.md#transaction-exchange)). Meanwhile the retry's nonce comes from `SimpleNonceManager`, which fills `eth_getTransactionCount(addr, pending)` ([alloy-provider 1.0.42, fillers/nonce.rs:43](https://docs.rs/alloy-provider/1.0.42/src/alloy_provider/fillers/nonce.rs.html)) -- so while the original sits in a pool at nonce N, the retry gets N+1 and the two txs are co-minable rather than mutually exclusive. (If the node dropped or never saw the original, pending count returns N, the retry reuses the same nonce, and the pair IS mutually exclusive -- the danger is exactly the accepted-but-errored case.)

To be precise about when a co-mined duplicate actually double-burns: the vault has no idempotency -- `redeem` -> `_withdraw` does `_burn(owner, shares)` + `receipt().managerBurn(...)` with balance checks only, no per-redemption state ([ethgild ReceiptVault.sol](https://github.com/rainlanguage/ethgild/blob/main/src/abstract/ReceiptVault.sol)). With a single redemption in flight the duplicate happens to revert: the first burn drained that redemption's shares, ERC-20 `_burn` fails on the duplicate, and the multicall is atomic. But with concurrent redemptions the bot wallet holds other redemptions' shares, and burn plans routinely consume receipts only partially -- every subcall of the duplicate then still has sufficient balance, so it lands and burns another redemption's backing. Two burns for one Alpaca journal, and the other redemption's own burn subsequently fails on insufficient shares. The contract only rejects the duplicate by accident of balance, not by design.

The nonce is also the clean fix: `SendableTxWithHash` already persists it. On an ambiguous broadcast failure, keep the persisted BurnIntended tx authoritative -- confirm or rebroadcast that exact raw tx -- and if a replacement is ever needed, re-sign it with the SAME nonce (bumping gas for replacement semantics). Protocol rules make same-nonce txs mutually exclusive: a tx is only valid when its nonce equals the account nonce, which increments on inclusion, so a given (account, nonce) can be consumed at most once ([execution-specs, `check_transaction` NonceMismatchError + `increment_nonce` in `process_transaction`](https://github.com/ethereum/execution-specs/blob/master/src/ethereum/forks/cancun/fork.py)). That makes the double-burn unrepresentable instead of dependent on wallet balances. Only prepare a fresh-nonce tx once the original nonce provably cannot land (account nonce advanced past it with no matching receipt).

src/redemption/mod.rs
```diff
@@ -1065,11 +1157,33 @@ impl EventSourced for Redemption {
...
+                    },
+                )
+                .await
+            }
         }
     }
 }
 
 impl Redemption {
+    #[allow(clippy::too_many_lines)]
```
0xgleb: should fix: this #[allow(clippy::too_many_lines)] was flagged in the previous round (the ask was to split apply_event into per-event helpers like apply_alpaca_called), the thread is marked resolved, but the attribute is still here and this PR is what added it. AGENTS.md's lint policy forbids adding #[allow(...)] without explicit permission. Extract the remaining event arms into apply_<event> helpers and drop the attribute.

src/redemption/burn_manager.rs
```diff
@@ -1018,21 +1079,25 @@ impl BurnManager {
         Err(BurnManagerError::InsufficientBalance { required, available })
     }
 
+    #[allow(clippy::too_many_lines)]
```
0xgleb: should fix: same story as apply_event, the prior round asked to remove the #[allow(clippy::too_many_lines)] suppressions and the thread is resolved, but only recover_single_burning was refactored. execute_burn_and_record_result still carries the attribute, added by this PR. Extract the IntendBurn preparation, submit, retry, and confirm blocks into named helpers and delete the allow.

src/redemption/burn_manager.rs
```diff
@@ -1131,6 +1265,22 @@ impl BurnManager {
...
+                if sendable_tx.is_some() && retry_count < MAX_RETRIES {
+                    return Box::pin(self.execute_burn_and_record_result(
+                        issuer_request_id,
+                        vault,
+                        plan,
+                        external_tx_id,
+                        retry_count + 1,
+                    ))
+                    .await;
+                }
```
0xgleb: should fix: this retry branch has zero test coverage. The submit-failure mock (new_submit_failure) never configures a prepared_tx, so sendable_tx is always None in every existing test and the recursion is unreachable. Given it re-executes the whole burn flow up to MAX_RETRIES times, it needs a test where prepare_burn_tx returns Some and submit_burn fails N times then succeeds, asserting exactly one successful submission and a settled reservation, plus an always-fails case asserting the aggregate ends Failed after MAX_RETRIES.

src/redemption/mod.rs
```diff
@@ -446,12 +465,19 @@ impl Redemption {
...
-            });
+        let stored_tx_id: String = match self {
+            Self::BurnSubmitted { fireblocks_tx_id, .. } => {
+                fireblocks_tx_id.clone()
+            }
+            Self::BurnIntended { sendable_tx: Some(tx), .. } => {
+                tx.hash.to_string()
+            }
+            _ => {
+                return Err(RedemptionError::InvalidState {
```
0xgleb, line 470: should fix: ConfirmBurn from BurnIntended { sendable_tx: Some(tx) } has no aggregate-level unit test, and the tx-id-mismatch sub-path for this new source state is completely untested. The only coverage goes through the BurnManager with a mock that never validates tx ids. Add a Given-When-Then test that replays to BurnIntended and confirms with the matching hash (expect TokensBurned), plus one with a mismatched id (expect the mismatch error).

src/redemption/burn_manager.rs
```diff
@@ -628,100 +724,56 @@ impl BurnManager {
...
-                    )
-                    .await;
-
-                match confirm_result {
-                    Ok(()) => {
-                        info!(target: "redemption", issuer_request_id = %issuer_request_id,
-                            "Burn confirmed successfully during recovery"
-                        );
+                .await
+            }
```
0xgleb, line 750: should fix: recovery for BurnIntended { sendable_tx: Some(..) } only polls ConfirmBurn by hash. If the process crashed after persisting BurnIntended but before broadcasting, the stored raw tx is never sent, so the redemption can only fail. The PR description says recovery will discard the stale tx and create a new one, but no path does either. Recovery should check the tx state and rebroadcast the persisted raw tx when it is absent, keeping the redemption recoverable on lookup timeouts.

src/vault/service.rs
```diff
@@ -213,150 +241,59 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
 
         let tx_hash: B256 =
             fireblocks_tx_id.parse().map_err(|_| VaultError::InvalidReceipt)?;
 
-        let receipt = self
-            .provider
-            .get_transaction_receipt(tx_hash)
-            .await?
-            .ok_or(VaultError::InvalidReceipt)?;
+        let receipt = PendingTransactionBuilder::new(
```
0xgleb, line 292: should fix: confirm_burn treats the 120s receipt timeout as a burn failure, but a signed EVM tx never expires. A slow-to-mine burn (fee spike, congestion, and fees here are fixed at prepare time) cascades into RecordBurnFailure, then the BurnFailed recovery pass polls once more for 120s and MarkFailed discards the tx hash entirely. The tx can still mine minutes later, and manual Reprocess would then submit a second burn alongside it. A receipt-poll timeout is ambiguous, not terminal: keep the redemption in a pending state that keeps polling the persisted hash, and only declare the tx dead once its nonce provably cannot be consumed (e.g. account nonce advanced past it with no receipt).

src/redemption/mod.rs
```diff
@@ -224,6 +224,21 @@ pub(crate) enum Redemption {
         reason: String,
         closed_at: DateTime<Utc>,
     },
+    BurnIntended {
+        metadata: RedemptionMetadata,
```
0xgleb, line 225: should fix: RecordBurnFailure and ConfirmBurn were extended to accept this new BurnIntended state, but handle_force_complete_burn (and the matching gate in BurnManager::force_complete_burn) still only accepts Burning and BurnSubmitted. A redemption stuck in BurnIntended after a crash is exactly the burn-landed-but-never-recorded situation the force-complete endpoint exists for, yet the operator gets InvalidState. Add BurnIntended to both matches.

src/redemption/burn_manager.rs
```diff
@@ -3630,6 +3801,311 @@ mod tests {
         ));
     }
 
+    /// Tests recovery from `BurnIntended { sendable_tx: Some(tx) }` (lines 719-816):
```
0xgleb: should fix: the new tests cover BurnIntended recovery when ConfirmBurn succeeds and the sendable_tx: None case, but not the crash window the feature exists for: BurnIntended persisted with a signed tx that was never broadcast. In that scenario ConfirmBurn times out, RecordBurnFailure fires, and it is unclear the reservation is released. Add a test seeding BurnIntended with a reservation, a confirm_burn mock returning a not-found error, and assert the reservation is released and the state lands somewhere recoverable.

src/redemption/burn_manager.rs
```diff
@@ -607,6 +608,101 @@ impl BurnManager {
...
+                        error = %err,
+                        "BurnIntended confirmation failed during recovery"
+                    );
+                }
+
+                if release_reservation {
+                    self.release_reserved_burn(vault, issuer_request_id).await;
+                }
+
+                self.store
```
0xgleb, line 697: minor: these two arms handle the reservation in opposite ways with no stated policy -- this one releases before the RecordBurnFailure write (a failed write then leaves a non-terminal redemption with its reservation already gone; safe only because release fires on definitive failures, release_burn is idempotent, and the next recovery pass re-runs this arm), while the non-vault arm below retains the reservation. The submit path documents its policy explicitly (the "Release only on a DEFINITIVE failure..." comment in execute_burn_and_record_result); give these recovery arms the same treatment so the asymmetry reads as policy rather than a leak.

src/redemption/mod.rs
```diff
@@ -909,6 +995,12 @@ impl EventSourced for Redemption {
                     found: "Uninitialized".to_string(),
                 })
             }
+            RedemptionCommand::IntendBurn { .. } => {
+                Err(RedemptionError::InvalidState {
+                    expected: "IntendBurn".to_string(),
```
0xgleb: minor: this was asked in the previous round and the thread is resolved, but expected is still "IntendBurn". Every other initialize() arm names the expected state, so the error reads "expected IntendBurn, found Uninitialized", which is nonsensical. Change it to "Burning or BurnIntended" to match handle_intend_burn.

src/vault/mod.rs
```diff
@@ -280,6 +302,8 @@ pub(crate) struct MultiBurnParams {
     /// retries after a previously accepted Fireblocks transaction failed.
     #[serde(default)]
     pub(crate) external_tx_id: Option<BurnExternalTxId>,
+    /// Block number of the detected_tx_hash.
```
0xgleb: minor: detected_block_number is a leftover from the removed find_existing_burn log-scan design. It is populated at both construction sites and in every test but never read anywhere, and the serde derives keep dead_code from firing. Remove it (or wire it into a real consumer if a follow-up needs it).

src/vault/service.rs
```diff
@@ -410,6 +347,88 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
+
+    /// Prepares a signed tx that can be sent on-chain via eth_sendRawTransaction
+    async fn prepare_burn_tx(
+        &self,
+        params: &super::MultiBurnParams,
+    ) -> Result<Option<SendableTxWithHash>, VaultError> {
+        let vault_contract =
+            OffchainAssetReceiptVault::new(params.vault, &self.provider);
+
+        let needs_encoding = params.burns.iter().any(|b| {
```
0xgleb: nit: the single-letter closure parameter |b| here was flagged last round and the thread is resolved, but it is still |b|. Rename to |burn| per the naming guideline.

src/redemption/burn_manager.rs
```diff
@@ -607,6 +608,101 @@ impl BurnManager {
         }
     }
 
+    async fn recover_single_burning_shared_inner(
```
0xgleb, line 618: should fix: `recover_single_burning_shared_inner` taking `tx_id: &String` is a symptom of the burn tx identity being stringly typed end-to-end, which docs/ids.md rule 1 calls a bug outright -- and this exact class caused a production bug before (`IssuerRedemptionRequestId` as `String`, fixed in PR #161 / RAI-631; see the "Why this matters" section of docs/ids.md).

The round-trip: `prepare_burn_tx` computes a typed hash (`SendableTxWithHash.hash: FixedBytes<32>`), `submit_burn` collapses it into `SubmittedTx.fireblocks_tx_id` via `.to_string()` (src/vault/service.rs), `BurnSubmitted` persists it as `fireblocks_tx_id: String`, `handle_confirm_burn` compares tx identity as *strings* (`stored_tx_id: String` vs `tx.hash.to_string()`), and `confirm_burn` re-parses our own authored value with `fireblocks_tx_id.parse().map_err(|_| VaultError::InvalidReceipt)` -- a fallible runtime parse with a misleading error variant. On top of that, the same `String` field carries a semantically different id on the Fireblocks backend (a Fireblocks tx id, not an EVM hash), so one field holds two id kinds distinguishable only by runtime parsing -- the exact "two ADTs collapsing into one `String` edge" failure docs/ids.md documents, and string equality on hex is also case-fragile where `B256` equality is 32 bytes.

Fix per ids.md rules 1-3: carry the alloy path's tx identity as `B256`/`TxHash` end-to-end (`BurnIntended` already stores it typed -- `BurnSubmitted` state and the confirm path should too), and where the two backends share a slot, make it one enum (`FireblocksTxId(String) | EvmTxHash(B256)`) rather than a shared `String`. Since `BurnSubmitted` pre-exists on main, keep its wire representation and parse once at the event boundary (the tolerant-deserialization pattern already used in this repo), typed everywhere past apply. The `&String` parameter disappears as a side effect.

src/vault/mod.rs
```diff
@@ -124,6 +126,26 @@ pub(crate) trait VaultService: Send + Sync {
...
+    /// but options return will be removed once fireblocks is completely removed
+    async fn prepare_burn_tx(
+        &self,
+        _params: &MultiBurnParams,
+    ) -> Result<Option<SendableTxWithHash>, VaultError> {
+        Ok(None)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
0xgleb, line 150: nit: SendableTxWithHash.hash is typed FixedBytes<32> while every other tx-hash field in this file (MultiBurnParams.detected_tx_hash, BurnVerification, the VaultError variants) uses the B256 alias for the identical type. Use B256 here too.

src/vault/service.rs
```diff
@@ -1,70 +1,98 @@
...
+                NonceFiller<SimpleNonceManager>,
+            >,
+            ChainIdFiller,
+        >,
+        WalletFiller<EthereumWallet>,
+    >,
+    RootProvider,
+>;
+
 /// Alloy-based blockchain service that interacts with the Rain OffchainAssetReceiptVault
```
0xgleb, line 43: nit: three claims in this doc block don't match the code: the struct is no longer generic over the provider, the precomputed hash is not deterministic from the burn request (it depends on nonce and gas fetched at prepare time), and recovery never replays the prepared tx, it only polls the stored hash. Worth rewording so the doc doesn't mislead the next reader.

src/redemption/mod.rs
```diff
@@ -497,9 +523,14 @@ impl Redemption {
         fireblocks_tx_id: Option<String>,
         planned_burns: Vec<BurnRecord>,
     ) -> Result<Vec<RedemptionEvent>, RedemptionError> {
-        if !matches!(self, Self::Burning { .. } | Self::BurnSubmitted { .. }) {
+        if !matches!(
+            self,
+            Self::Burning { .. }
+                | Self::BurnIntended { .. }
```
0xgleb, line 532: should fix: the `handle_record_burn_failure` guard now accepts `BurnIntended` as a source state, but no aggregate-level unit test exercises that transition. This is the path taken when prepare fails or submit exhausts retries. Add a test that replays to `BurnIntended`, issues `RecordBurnFailure`, and asserts `BurningFailed` is emitted and the aggregate ends `Failed`.

JuaniRios (approved):

src/redemption/cmd.rs
0xgleb: Addressed by replacing the scan/retry boolean flow entirely. Burn intent is now a persisted state, recovery uses a typed BurnRecoveryState internally, and no call site passes a bare recovery boolean.

src/redemption/mod.rs
0xgleb: Addressed in the replacement recovery flow. Recovery logs the issuer request ID and persisted transaction hash before exact-byte re-broadcast, then logs successful confirmation.

src/redemption/burn_manager.rs
0xgleb: Addressed with BurnManager regressions covering exact persisted-transaction re-broadcast and confirmation, ambiguous broadcast retention, and confirmation timeout retention.

src/redemption/mod.rs
0xgleb: Superseded by the crash-safe intent design. There is no best-effort event scan followed by resubmission now; an ambiguous broadcast keeps BurnIntended, the exact signed transaction, and its reservation for recovery.

src/vault/service.rs
0xgleb: Addressed. The service docs now describe persistence of the complete signed transaction and assigned nonce, exact-byte re-broadcast on recovery, and confirmation by the stored hash.

src/vault/service.rs
0xgleb: Superseded: the recovery log scan and its block batching were removed. Recovery uses the persisted signed transaction and transaction hash instead.

src/vault/service.rs
0xgleb: Addressed. SubmittedTx is generic over the external ID type, so BurnExternalTxId remains typed across the internal boundary without a String round-trip.

src/redemption/mod.rs
0xgleb: Superseded by the persisted-intent flow and covered at the new boundaries: tests now cover exact re-broadcast and completion, ambiguous broadcast retention, and pending confirmation retention with log assertions.

src/vault/service.rs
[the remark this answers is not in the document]
0xgleb: Superseded. Recovery no longer infers identity from any Withdraw in a multicall. It persists and re-broadcasts the exact signed transaction bytes and confirms that transaction hash.

src/redemption/burn_manager.rs
0xgleb: Addressed. Ambiguous broadcast no longer recurses into a fresh-nonce burn. BurnIntended persists the exact signed bytes, hash, and nonce; recovery re-broadcasts that transaction and keeps the reservation until a definitive outcome.

src/redemption/mod.rs
0xgleb: Addressed. The apply_event suppression was removed and event handling is split into named per-event helpers.

src/redemption/burn_manager.rs
0xgleb: Addressed. The execute_burn_and_record_result suppression was removed and the preparation, submission, and confirmation paths are split into named helpers.

src/redemption/burn_manager.rs
0xgleb: Superseded with the unsafe recursive retry branch itself. Ambiguous submission now retains BurnIntended and the exact signed transaction; regressions cover ambiguous retention, definitive revert handling, and exact re-broadcast recovery.

src/redemption/mod.rs
0xgleb: Addressed with aggregate tests for ConfirmBurn from BurnIntended using both the matching stored hash and a mismatched hash.

src/redemption/burn_manager.rs
0xgleb: Addressed. BurnIntended recovery re-broadcasts the exact persisted raw transaction once before confirming its stored hash; the crash-before-broadcast path has a manager-level regression.

src/vault/service.rs
0xgleb: Addressed. Receipt timeout is typed as ConfirmationPending and remains in the submitted recoverable state with the same transaction ID and held reservation; recovery continues polling instead of submitting a replacement.

src/redemption/mod.rs
0xgleb: Addressed. ForceCompleteBurn accepts BurnIntended in both the aggregate and BurnManager gate, with aggregate and endpoint coverage.

src/redemption/burn_manager.rs
0xgleb: Addressed under the safer recovery policy: the persisted transaction is re-broadcast exactly, and ambiguous confirmation keeps the intent/submission recoverable with the reservation held. Tests cover exact re-broadcast and pending confirmation retention.

src/redemption/burn_manager.rs
0xgleb: Addressed. The recovery code now documents the policy at each branch: release only for definitive no-burn outcomes, retain for ambiguous outcomes, and settle after confirmed completion.

src/redemption/mod.rs
0xgleb: Addressed. The invalid-state expectation now reads Burning or BurnIntended.

src/vault/mod.rs
0xgleb: Addressed. The unused detected_block_number field and all construction-site plumbing were removed.

src/vault/service.rs
0xgleb: Addressed. The closure argument is named burn.

src/redemption/burn_manager.rs
0xgleb: Addressed. On-chain hashes remain typed through TxId backed by B256, while legacy Fireblocks identifiers occupy the distinct legacy variant; no internal String parse round-trip remains.

src/vault/mod.rs
0xgleb: Addressed. SendableTxWithHash uses the B256 alias consistently.

src/vault/service.rs
0xgleb: Addressed. The documentation now describes signed transaction persistence, nonce assignment, exact-byte re-broadcast, and stored-hash confirmation without the obsolete generic/deterministic claims.

src/redemption/mod.rs
0xgleb: Addressed with an aggregate regression that replays BurnIntended, records the failure, and asserts BurningFailed is emitted.

0xgleb (approved):

src/admin.rs
```diff
@@ -644,7 +646,9 @@ async fn inspect_prior_fireblocks_burn(
                     issuer_request_id,
                     RedemptionCommand::RecordExistingBurn {
                         issuer_request_id: issuer_request_id.clone(),
-                        fireblocks_tx_id: fb_tx_id.clone(),
+                        fireblocks_tx_id: crate::vault::TxId::Legacy(
+                            fb_tx_id.clone(),
+                        ),
```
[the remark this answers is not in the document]
0xgleb: Addressed in ff3ab7c: `TxId` is imported through the module-level vault import blocks and the affected production/test sites now use the bare type consistently.

tests/mint.rs
```diff
@@ -76,6 +77,11 @@ async fn perform_mint_flow(
     let user_signer = PrivateKeySigner::from_bytes(&user_private_key)?;
     let user_wallet_instance = EthereumWallet::from(user_signer);
     let user_provider = ProviderBuilder::new()
+        .disable_recommended_fillers()
+        .with_gas_estimation()
+        .filler(BlobGasFiller)
+        .with_simple_nonce_management()
+        .filler(ChainIdFiller::default())
```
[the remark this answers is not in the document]
0xgleb: Addressed in ff3ab7c: the mint integration test now uses the shared `create_provider` helper; the duplicated filler-chain imports and setup were removed.

SPEC.md
```diff
@@ -1696,10 +1707,13 @@ stateDiagram-v2
...
     AlpacaCalled --> Failed: RecordAlpacaFailure / MarkFailed
     Burning --> Completed: RecordBurnSuccess
-    Burning --> BurnSubmitted: BurnTokens (BurnFireblocksSubmitted)
+    Burning --> BurnIntended: IntendBurn
     Burning --> Failed: RecordBurnFailure / MarkFailed
     Burning --> Completed: ForceCompleteBurn (admin, verified on-chain)
     Burning --> Closed: CloseRedemption (admin)
+    BurnIntended --> BurnSubmitted: BurnTokens (BurnFireblocksSubmitted)
+    BurnIntended --> Failed: RecordBurnFailure / MarkFailed
+    BurnIntended --> Completed: ForceCompleteBurn (admin, verified on-chain)
```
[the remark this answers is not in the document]
0xgleb: Addressed in ff3ab7c: the state diagram now documents the direct `BurnIntended -> Completed` recovery transition through `ConfirmBurn`, separately from the verified admin force-complete path.

src/redemption/mod.rs
```diff
@@ -224,6 +221,21 @@ pub(crate) enum Redemption {
...
+        /// Dust quantity to return to user
+        dust_quantity: Quantity,
+        called_at: DateTime<Utc>,
+        alpaca_journal_completed_at: DateTime<Utc>,
+        planned_burns: Vec<event::BurnRecord>,
+        #[serde(default)]
+        external_tx_id: Option<BurnExternalTxId>,
+        #[serde(default)]
+        sendable_tx: Option<SendableTxWithHash>,
+    },
```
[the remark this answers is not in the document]
0xgleb: Addressed in ff3ab7c: `BurnIntended` now participates in both shared accessors, so its metadata and Alpaca quantity remain available through the generic redemption API.

src/vault/service.rs
```diff
@@ -410,6 +343,87 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
+        let hash = *envelop.tx_hash();
+        let tx = envelop.encoded_2718();
+        return Ok(Some(SendableTxWithHash {
+            tx,
+            hash,
+            nonce,
+            signed_at: Utc::now(),
+            dust_shares: params.dust_shares,
+        }));
+    }
```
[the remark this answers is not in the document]
0xgleb: Addressed fully in abbd2cca. The shared build_burn_calls helper now owns receipt metadata fallback encoding and redeem calldata construction; both the Alloy and Fireblocks paths call it, so those paths can no longer drift. I also retained the earlier typo and needless-return cleanups from this compound comment.

0xgleb:
Follow-up review fixes are in abbd2cca: (1) extracted shared burn calldata construction used by both Alloy and Fireblocks, completing the previously only-partially-addressed compound review comment; (2) changed recovery to classify aggregate state after a successful step instead of issuing an expected-invalid extra command, with assertions that terminal recovery emits no domain-error log; and (3) fixed the redemption-dust regression test so each scenario returns its actual expected quantity, asserts the bot balance was burned, and allows enough time for the production 5-second poll interval.

src/redemption/burn_manager.rs
```diff
@@ -607,6 +610,107 @@ impl BurnManager {
...
+                    .send(
+                        issuer_request_id,
+                        RedemptionCommand::RecordBurnFailure {
+                            issuer_request_id: issuer_request_id.clone(),
+                            error: err.clone(),
+                            fireblocks_tx_id: Some(tx_id.clone()),
+                            planned_burns: planned_burns.to_vec(),
+                        },
+                    )
+                    .await?;
```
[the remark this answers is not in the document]
0xgleb: Addressed in dcf043dd. Reservation-release failures now propagate instead of being logged and swallowed, so the aggregate is terminalized only after the idempotent release succeeds. The new regression test forces release to fail after prepare-tx failure and verifies that the reservation remains and the aggregate stays recoverable in Redemption::Burning.

src/redemption/burn_manager.rs
```diff
@@ -1334,20 +1663,21 @@ const fn aggregate_state_name(aggregate: &Redemption) -> &'static str {
...
+pub(crate) fn extract_fireblocks_tx_id(error: &VaultError) -> Option<TxId> {
     match error {
         VaultError::Fireblocks(
             crate::fireblocks::FireblocksVaultError::TransactionFailed {
                 tx_id,
                 ..
             }
             | crate::fireblocks::FireblocksVaultError::MissingTxHash { tx_id },
-        ) => Some(tx_id.clone()),
+        ) => Some(TxId::Legacy(tx_id.clone())),
```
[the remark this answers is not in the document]
0xgleb: Addressed in dcf043dd. I imported BurnExternalTxId, TxId, FireblocksVaultError, and MultiBurnEntry at their use sites and removed the flagged fully-qualified internal paths across src/mint/mod.rs, src/mint/recovery.rs, and src/redemption/burn_manager.rs.

src/vault/mock.rs
```diff
@@ -95,13 +104,16 @@ pub(crate) struct MockVaultService {
     /// verification, exercising the admin force-complete happy path.
     #[cfg(test)]
     verify_burn: Arc<Mutex<MockVerifyBurn>>,
+    /// Signed tx returned by `prepare_tx` when local signing is configured.
+    /// `None` simulates the Fireblocks backend (prepare_tx returns nothing).
+    #[cfg(test)]
+    prepared_tx: Arc<Mutex<Option<SendableTxWithHash>>>,
```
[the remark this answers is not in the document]
0xgleb: Addressed in dcf043dd. MockVaultService::reset now clears prepared_tx along with the other transaction state. The new reset_clears_prepared_transaction regression test prepares a transaction, resets the mock, and verifies that the prepared transaction is gone.

src/vault/service.rs
```diff
@@ -1,70 +1,102 @@
...
+            JoinFill<
+                JoinFill<JoinFill<Identity, GasFiller>, BlobGasFiller>,
+                NonceFiller<SimpleNonceManager>,
+            >,
+            ChainIdFiller,
+        >,
+        WalletFiller<EthereumWallet>,
+    >,
+    RootProvider,
+>;
```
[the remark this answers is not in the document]
0xgleb: Addressed in dcf043dd. RealBlockchainServiceProvider is an internal wiring alias, so its visibility is now pub(crate) rather than public API.

0xgleb:
Follow-up from the final CI pass: GitHub reported that the pinned checkout and Nix-cache actions still targeted deprecated Node 20. Updated checkout to SHA-pinned v7.0.0 and cache-nix-action to SHA-pinned v7. While touching the workflows, added explicit read-only permissions and job timeouts, and moved deploy inputs/secrets out of inline shell interpolation. The full stack has been restacked and pushed.

PR #216, opened 2026-07-04
docs: spec orchestrator-based mint/burn flows

JuaniRios:
## Motivation

Today the issuance bot mints and burns tokens by talking to the vault contracts directly, and it does all the receipt bookkeeping itself in its own database. A new contract, the ST0x Orchestrator, is going to take that job over. It becomes the single place every mint and burn goes through, it holds the receipts, and it does the bookkeeping on-chain instead of in the bot.

That makes the bot simpler in the long run, but a lot of moving parts have to change to point at the orchestrator instead of the vault. Our rule in this repo is that we write the plan in SPEC.md before writing any code. This PR is only that plan. The actual code changes are separate tickets, starting at [RAI-1217](https://linear.app/makeitrain/issue/RAI-1217).

## Solution

Docs only. This edits SPEC.md and changes no code. It writes down how minting and burning work once they go through the orchestrator, and it settles seven design decisions so the follow-up tickets have concrete answers instead of open questions. The decisions now live in the spec itself (Orchestrator Migration -> Design Decisions), so you do not have to go dig them out of Linear.

The seven decisions, in short:

1. **Mint approval (resolved):** the wallet receiving the tokens has to sign each mint. The liquidity bot owns that wallet, so it signs the mint itself and hands the signature to the issuance bot over the internal connection the two bots already use (the same one that checks whether an asset is frozen). Nothing has to travel through Alpaca, which is what makes this simple.
2. **Retry safety:** each mint gets one fixed number (a nonce) that we save before submitting and reuse on every retry, so "this number was already used" means "this mint already happened." Recovery then double-checks the on-chain mint matches the exact token and amount, not just the number.
3. **Code shape:** the existing signing interface gets two new methods for the orchestrator path, and a per-asset setting in a new TOML config file (the same pattern the liquidity bot uses; default: old path for every asset) picks which path each asset uses, so we can ship it turned off and roll back easily.
4. **New records:** five new events for the orchestrator path. Existing events stay untouched except for one new optional field. Failures reuse the machinery we already have, with a typed reason attached.
5. **Token approval:** approve each token once, up front, as an ops step, plus a check right before burning so a missing approval fails with a clear error instead of a confusing revert.
6. **Dust (settled):** the sub-nanotoken crumb left over after a burn stays in the bot wallet and gets recorded, rather than being sent back. Returning it would cost a whole separate transaction per redemption for less than one billionth of a token, so we keep it.
7. **Per-asset rollout:** the orchestrator is switched on one asset at a time, not all at once. We can pilot it on a single low-volume asset in production while every other asset keeps the proven old path, and if something goes wrong we roll back just that one asset. The liquidity bot learns which assets need the new mint signature from the asset-status endpoint it already calls, so the two bots cannot disagree mid-rollout.

The reasoning behind each decision, and the rules for every failure case, are in the spec's Design Decisions and Failure States sections. The migration ticket breakdown, meaning who does what and in what order, stays in Linear on [RAI-1215](https://linear.app/makeitrain/issue/RAI-1215).

## Anything else

This spec was hardened with several automated review passes, but the last round of edits was not independently re-reviewed, so it is worth a careful human read before the follow-up tickets build on it. The trickiest part to check is the mint recovery rule in decision 2: when a retry reports the mint was already done, we must confirm it matches the exact token and amount we meant, not just the wallet and a reused number, or we could mark the wrong mint as complete.

Sequencing note: this whole orchestrator change is meant to ship after the issuer wallet moves from Fireblocks to Turnkey ([RAI-1123](https://linear.app/makeitrain/issue/RAI-1123)), so it lands once Turnkey is live and Fireblocks is gone. The plan is signer-agnostic, so the parts that mention Fireblocks just mean whatever signer is active at the time, which will be Turnkey.

## Checks

## Summary by CodeRabbit

* **New Features**
  * Added full orchestrator-based custody, mint, and burn support alongside the existing vault-direct flow.
  * Status responses now include the active asset vault mode.
  * Documented orchestrator configuration via TOML and added guidance for an admin “close mint” action.
  * Updated user-facing mint/redemption updates and views to reflect orchestrator-mode events.
* **Bug Fixes**
  * Improved mode consistency across confirm/recovery paths by using the stored mode from the original operation.
  * Strengthened protections against duplicate/replay and tightened failure reporting, including orchestrator-specific dust handling.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -71,6 +71,25 @@ settlement layer between Authorized Participants (APs) and us.
 - On-chain address where APs send tokens to redeem
 - We monitor this address for incoming transfers
 
+**ST0xOrchestrator Contract:**
+
+- A singleton contract that holds custody of **all** ERC-1155 receipts across
+  every vault, replacing the bot-wallet receipt custody model for the vaults it
+  covers. Exposes role-gated `mint()`/`burn()` entry points guarded by
+  `MINT_ROLE`, `BURN_ROLE`, and an `EMERGENCY_ROLE` for manual recovery actions
+  (e.g. moving receipts, adjusting the burn pointer).
+- **Dual-mode, config-selected per asset, default vault-direct.** Each asset
+  operates in `vault_direct` mode (today's direct `OffchainAssetReceiptVault`
+  multicall flow, described above and in "Complete Mint Flow" / "Complete
+  Redemption Flow") or `orchestrator` mode (this contract). The mode is a
+  per-asset configuration choice (`[assets.<UNDERLYING>]` in the TOML config
+  file, see "Configuration" -> "TOML Configuration File"), not a per-request
+  choice. It defaults to `vault_direct` for every asset, so the orchestrator can
+  be dark-deployed and exercised before cutover; cutover moves assets into
+  orchestrator mode incrementally — a single pilot asset first, then the rest —
+  and rollback flips a single asset back. See "Orchestrator Migration
+  (ST0xOrchestrator)" for the full design.
+
 ### ES/CQRS Architecture
 
 The issuance bot uses **Event Sourcing (ES)** and **Command Query Responsibility
@@ -174,8 +193,18 @@ initial request through journal confirmation to on-chain minting and callback.
 - `tokenization_request_id`: Alpaca's identifier
 - `quantity`, `underlying`, `token`, `network`, `client_id`, `wallet`: Request
   details
+- `mint_mode: VaultMode`: the asset's resolved `VaultMode` at initiate time.
+  Anchors mode-derivation for this mint — see "Orchestrator Migration" ->
+  "Recipient Authorization"
+- `mint_authorization` (orchestrator mode only): the `MintAuthV1` the liquidity
+  bot supplied for this mint via the internal mint-authorization call, absent
+  until that call arrives — see "Orchestrator Migration" -> "Recipient
+  Authorization"
 - `status`: Current state in the mint lifecycle
-- `tx_hash`, `receipt_id`, `shares_minted`: On-chain transaction details
+- `tx_hash`, `receipt_id`, `shares_minted`: On-chain transaction details.
+  Orchestrator-mode mints populate the analogous on-chain proof (`nonce` instead
+  of `receipt_id`) — lifecycle state names stay backend-agnostic; only the
+  audit-data shape differs
 - Timestamps for each lifecycle stage
 
 **Durable jobs vs pure commands:** On-chain I/O and Alpaca callbacks are
@@ -193,6 +222,13 @@ plus enqueue of the next job), not a wallet-locked aggregate command.
   transfer
 - `RejectJournal { issuer_request_id, reason }` - Alpaca rejected shares journal
   transfer
+- `AuthorizeMint { issuer_request_id, mint_authorization }` (orchestrator mode
+  only) - Associate the liquidity bot's `MintAuthV1` with this mint, delivered
+  out-of-band via the internal mint-authorization call after `Initiate`.
+  Validates the EIP-712 signature and nonce (see "Recipient Authorization"),
+  rejects delivery when this mint's `mint_mode` is `VaultDirect`, and is
+  idempotent on redelivery of an identical authorization. Produces
+  `MintAuthorizationReceived`. Does not change the lifecycle state
 - `Deposit { issuer_request_id }` - Record intent to mint. Pure state transition
   from `JournalConfirmed` to `Minting` — no network call. Produces
   `MintingStarted`. Intent is persisted before the network call so that a crash
@@ -270,7 +306,19 @@ authoritative statement of what each observation records.
 
 **Events:**
 
-- `Initiated` - Mint request created (carries all request details)
+- `Initiated` - Mint request created (carries all request details). Gains one
+  additive optional `mint_mode` field (`#[serde(default)]`, `VaultMode`,
+  defaulting to `VaultDirect` for historical events, which all predate
+  orchestrator mode and so could only have been vault-direct) recording the
+  asset's resolved `VaultMode` at initiate time, before any possible mint
+  submission; this anchors mode-derivation for the mint the same way
+  `RedemptionDetected.burn_mode` anchors it for Redemption
+- `MintAuthorizationReceived` (orchestrator mode only) - The liquidity bot's
+  validated `MintAuthV1` for this mint, delivered via the internal
+  mint-authorization call. Carries
+  `{issuer_request_id, mint_authorization, received_at}`. This is the
+  persistence point for the nonce — `Initiated` is written on the Alpaca POST,
+  strictly before the authorization exists, so it cannot carry one
 - `JournalConfirmed` - Alpaca journal transfer confirmed
 - `JournalRejected` - Alpaca journal transfer rejected (terminal)
 - `MintingStarted` - Mint intent recorded (aggregate moves to `Minting`)
@@ -279,7 +327,13 @@ authoritative statement of what each observation records.
 - `MintTxSubmitted` - Persisted signed mint transaction broadcast (carries
   `external_tx_id` and `tx_id` — the on-chain tx hash — for crash recovery)
 - `TokensMinted` - On-chain mint succeeded (carries tx details)
-- `MintingFailed` - On-chain mint failed
+- `MintingFailed` - On-chain mint failed. Gains one additive optional
+  `classification` field (`#[serde(default)]`, `MintFailureClassification`,
+  default `Unclassified`) — see "Orchestrator Migration" -> "Failure States" for
+  the decodable on-chain-revert variants (and "Recipient Authorization" ->
+  "Nonce" for `NonceConsumedByOtherMint` and `NonceReplayUnresolved`, both
+  assigned by recovery's own full-match check rather than decoded from a revert)
+  and how retry-exclusion and logging key off it
 - `MintCompleted` - Alpaca callback sent, mint fully completed (terminal)
 - `ExistingMintRecovered` - Existing on-chain mint discovered during recovery
   (carries tx details)
@@ -288,6 +342,27 @@ authoritative statement of what each observation records.
   correlating the command with the event it commits, so queue dispatch can
   distinguish a successful transition from an idempotent no-op against
   already-advanced state
+- `MintClosed` - Admin-closed mint (terminal). Carries the operator `reason`,
+  `closed_at`, and optional `acknowledged_unresolved_mint_nonce` (present only
+  when closing a `NonceReplayUnresolved` mint). Closed mints do not appear in
+  stuck queries
+- `OrchestratorTokensMinted` (orchestrator mode only) - On-chain orchestrator
+  mint succeeded. Carries
+  `{issuer_request_id, tx_hash, nonce, shares_minted,
+  gas_used, block_number, minted_at}`
+  — nonce replaces `receipt_id` since the orchestrator, not the bot, owns
+  receipt custody
+- `OrchestratorMintRecovered` (orchestrator mode only) - Existing on-chain
+  orchestrator mint discovered during recovery, via a proactive `Minted`-log
+  query keyed on `(wallet, nonce)` and confirmed by an exact match on this
+  mint's own `token` and `amount` (see "Nonce" below — the nonce-uniqueness view
+  is `(wallet, nonce)` only, so a bare hit is not sufficient proof) — mirroring
+  vault-direct recovery's proactive receipt-inventory check
+  (`RecordExistingMint`, see "Recovery orchestration" above). A `NonceReplayed`
+  revert is only the fallback signal for a submit/query race, not the primary
+  discovery path. Carries
+  `{issuer_request_id, tx_hash, nonce, shares_minted, block_number,
+  recovered_at}`
 
 Newly persisted transaction IDs use an explicitly tagged `hash` or `legacy`
 representation so replay preserves the original `TxId` variant, including a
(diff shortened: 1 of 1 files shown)


JuaniRios:
* **#216** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/216?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/216?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=216).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

SPEC.md
```diff
@@ -288,6 +358,33 @@ receipt for a mint in `MintingFailed` state (where the predecessor was
...
+"Dual-Mode Operation and Cutover"), so `RecoverFromReceipt` does not apply;
+`Recover`'s existing-mint check instead queries the orchestrator's `Minted` log
+by `(wallet, nonce)`, emitting `OrchestratorMintRecovered` only when the log's
+`token`/`amount` also exactly match this mint's own request facts (see "Nonce"
+below for the full-match rule and its manual-failure fallback). See
+"Orchestrator Migration" for the mint flow, failure states
+(`BadRecipientSignature`, `RecipientCallbackRejected`, `VaultAmountMismatch`,
+`VaultLogicMismatch`/`ReceiptLogicMismatch`), and the full event reuse/new
+rationale.
+
```
0xgleb, line 594: should fix: mode derivation keys off mint_authorization on Initiated, but Initiated is written on the Alpaca POST before the liquidity bot delivers authorization on the internal call (see the sequence diagram and Decision 1). The spec also allows associating auth via a separate step, which cannot update an already-persisted Initiated in ES. An orchestrator mint will look vault-direct at recovery time. Mirror redemption: persist an explicit mint_mode on Initiated from config at initiate time, and treat mint_authorization as orthogonal data that may arrive later.

SPEC.md
```diff
@@ -560,6 +776,700 @@ the CQRS command through the `Store` (never writing the `events` table
...
+**If a recipient other than the liquidity bot is ever needed** (e.g. a future
+non-liquidity-bot AP), the fallback is the bot signing its own authorization (a
+weaker security property: the bot's key becomes sufficient to both mint and
+authorize) or a dedicated `IMintRecipient` callback contract that approves
+mints. See Decision 1 in "Design Decisions" above for the full alternative
+analysis.
+
+**Nonce.** The nonce is fixed per mint operation, chosen by whichever party
+produces the signature — today, the liquidity bot, as recipient/signer, under
+`MintAuthV1` — persisted on `Initiated` before the first on-chain submission,
```
0xgleb: should fix: this says the nonce is persisted on Initiated before the first on-chain submission, but the nonce is chosen by the liquidity bot and arrives with authorization after Initiate. Same root issue as the mode anchor above. Either Initiated cannot be the persistence point for nonce, or you need a follow-up event and updated recovery rules.

SPEC.md
```diff
@@ -560,6 +776,700 @@ the CQRS command through the `Store` (never writing the `events` table
...
+- Views: `nonceUsed(address to, bytes32 nonce) -> bool`,
+  `vaultLogicIsExpected() -> bool`,
+  `nextBurnReceiptId(address token) ->
+  uint256`.
+- Events: `Minted(token, to, amount, nonce, ...)`,
+  `Burned(token, amount, burn_range,
+  ...)` — per `IST0xOrchestratorV1.sol`
+  (st0x.deploy PR #222) this is
+  `Burned(caller, token, amount, firstReceiptId, nextBurnReceiptIdAfter)`;
+  `burn_range: (start_id, end_id)` below maps to that pointer pair. The range
```
0xgleb: should fix: burn_range is called out as the audit record for orchestrator burns, but the text admits the boundary receipt may be only partially consumed and says to confirm against the final ABI later. For a planning doc that downstream tickets will implement against, I'd pin the exact mapping from Burned(firstReceiptId, nextBurnReceiptIdAfter) now, including inclusive/exclusive bounds and what shares_burned proves vs pointer movement.

SPEC.md
```diff
@@ -560,6 +776,700 @@ the CQRS command through the `Store` (never writing the `events` table
...
+| `BurningFailed`                                                                                                                    | Reused, plus one additive optional `classification` field (`#[serde(default)]`, `BurnFailureClassification`) | Decodes `InsufficientReceipts`/`AllowanceInsufficient`/`ReceiptLogicMismatch`/`VaultLogicMismatch` (post-submit race) so retry-exclusion and admin grouping key off a typed field, never `error: String`; `planned_burns` is already `#[serde(default)]`-tolerant of an empty plan |
+| `BurnFireblocksSubmitted`                                                                                                          | New: `OrchestratorBurnSubmitted`                                                                             | Existing `planned_burns: Vec<BurnRecord>` is a required per-receipt plan; orchestrator mode has none                                                                                                                                                                               |
```
0xgleb: minor: the spec names this event BurnFireblocksSubmitted, but the codebase persists RedemptionEvent::BurnTxSubmitted (src/redemption/event.rs). The new orchestrator mapping builds on the spec name. Worth aligning before RAI-1220/1221 wire views and handlers.

SPEC.md
```diff
@@ -560,6 +776,700 @@ the CQRS command through the `Store` (never writing the `events` table
...
+| `RedemptionDetected`                                                                                                               | Reused, plus one additive optional `burn_mode` field (`#[serde(default)]`, `VaultMode`)                      | Anchors mode-derivation for this redemption at the earliest possible point, before any burn submission                                                                                                                                                                             |
+| `AlpacaCalled`, `AlpacaCallFailed`, `AlpacaJournalCompleted`, `RedemptionFailed`, `Reprocessed`, `RedemptionClosed`, `BurnResumed` | Reused unchanged                                                                                             | Pure Alpaca/lifecycle facts, pre-burn                                                                                                                                                                                                                                              |
```
0xgleb: nit: RedemptionFailed is newly documented in this PR events list and MarkFailed mapping, but the orchestrator table groups it under reused unchanged. Small labeling tweak would make the doc easier to scan.

0xgleb (changes requested):

SPEC.md
JuaniRios: Think this one's a misread, the code already matches the spec. `src/redemption/event.rs:171` defines `BurnFireblocksSubmitted`, and there's no `BurnTxSubmitted` variant anywhere in the crate (only `BurnFireblocksSubmitted`, used across `admin.rs`, `burn_manager.rs`, `mod.rs`, and `view.rs`).

The naming is also called out on purpose in SPEC.md:790-800: the `FireblocksSubmitted` / `BurnFireblocksSubmitted` names are kept unchanged for event-history compatibility even after we move off Fireblocks onto Turnkey, and everywhere the doc says "Fireblocks" you should read it as "the active signing backend".

So no rename needed before RAI-1220/1221. Lmk if you were thinking of a different event tho, happy to be wrong here.

SPEC.md
```diff
@@ -731,6 +978,733 @@ the CQRS command through the `Store` (never writing the `events` table
...
+
+  **RAI-1220 acceptance criterion.** This reading is the one the spec requires,
+  inferred from the field name `nextBurnReceiptIdAfter` and the
+  `nextBurnReceiptId(token)` view — it is **not yet verified against the merged
+  contract**. Before wiring the `Burned` decoder, assert against the merged
+  `IST0xOrchestratorV1.sol` that (a) `nextBurnReceiptIdAfter` is exclusive of
+  the receipts this burn fully drained, and (b) it names the receipt a
+  subsequent burn resumes from. Encode both in a decoder test against a real
+  `Burned` log. If the merged ABI disagrees, correct this section first — do not
+  work around it in the decoder.
```
rouzwelt, line 1668: **Verify the** **`Burned`** **range semantics before implementing the decoder.**

The half-open `burn_range` interpretation is critical for receipt-pointer accounting, but this section explicitly says it is not yet verified against the merged contract. Treating an incorrect endpoint as exclusive could skip or double-process receipts. Block decoder implementation until the ABI and a real-log test confirm the semantics.

SPEC.md
```diff
@@ -731,6 +978,733 @@ the CQRS command through the `Store` (never writing the `events` table
...
+mint with exactly the signed `(token, to, amount, nonce)` values, since the
+orchestrator verifies the signature against them exactly.
+
+The authorization is persisted by its own command and event — `AuthorizeMint` ->
+`MintAuthorizationReceived` (see "Mint Aggregate" above) — never on `Initiated`.
+`Initiated` is written synchronously on Alpaca's `POST /inkind/issuance`, which
+strictly precedes this internal call, and events are immutable once written. The
+exact internal-call wire shape and correlation mechanism remain owned by
+RAI-1220 (issuance side) plus the liquidity bot; the aggregate-level command,
+event, and ordering above are settled here.
```
rouzwelt: **Define one canonical authorization correlation key.**

The specification leaves the wire correlation as “e.g. by `tokenization_request_id`,” while `AuthorizeMint` is keyed by `issuer_request_id`. Make the key, uniqueness rules, and mismatch behavior explicit before implementation; an incorrect association could attach valid authorization to the wrong mint.

SPEC.md
```diff
@@ -234,10 +270,37 @@ initial request through journal confirmation to on-chain minting and callback.
...
+  (`TokensMinted`/`OrchestratorTokensMinted`/`ExistingMintRecovered`/
+  `OrchestratorMintRecovered`) is already recorded in this aggregate's own
+  history; an economically committed mint must be resolved through the success
+  path (`SendCallback`), never hidden by `CloseMint`. From the remaining,
+  pre-mint states, the handler additionally checks for a matching on-chain mint
+  before closing (vault-direct: no receipt found for this `issuer_request_id`;
+  orchestrator: no `Minted` log matches `(to, nonce, token, amount)` — see
+  "Nonce" below) — if one is found, `CloseMint` is rejected and the operator
+  runs `Recover` to reconcile the already-completed mint instead. Emits
+  `MintClosed`. Closed mints are excluded from recovery and stuck queries
```
rouzwelt: **Do not close mints with an unresolved submitted transaction.**

Checking that no matching on-chain mint exists at one instant is insufficient for `TxSubmitted` or `MintingFailed` states: the persisted transaction may still be pending and land after `MintClosed`, while recovery and callback processing are excluded. Require a confirmed revert or an equivalent proof that the transaction cannot land before allowing closure.

SPEC.md
```diff
@@ -1312,6 +2379,10 @@ in Step 2, we'll find out in Step 3.
   - "Insufficient Eligibility: Client not eligible"
   - "Failed Validation: Invalid data payload"
 
+  (Orchestrator mode only: a missing/invalid `mint_authorization` or an
+  already-used nonce is rejected by the internal mint-authorization call, not
+  this endpoint — see "Orchestrator Migration" -> "Recipient Authorization".)
```
rouzwelt: **Distinguish missing authorization from invalid authorization.**

A missing authorization cannot be “rejected by the internal call” if no call arrives; it leaves the mint waiting or blocked before submission. Document that behavior separately from invalid signatures and reused nonces so callers and operators know whether to retry, wait, or remediate.

SPEC.md
```diff
@@ -731,6 +978,733 @@ the CQRS command through the `Store` (never writing the `events` table
...
+risk a mis-backed issuance. It is instead recorded as `MintingFailed` with
+`classification:
+MintFailureClassification::NonceConsumedByOtherMint`, a
+non-retryable, manual-intervention failure. The nonce can never be reused for
+this mint either way, so — exactly like `BadRecipientSignature` below — the only
+resolution is `CloseMint` on the stranded (already-journaled) aggregate followed
+by a brand-new `Initiate` paired with fresh authorization (and a fresh nonce)
+from the liquidity bot via the internal mint-authorization call. `ConfirmMint`
+applies this same full-match check when it is the one to observe a
+`NonceReplayed` revert on a submitted transaction, rather than `Recover`.
```
rouzwelt: **Do not classify a missing log as** **`NonceConsumedByOtherMint`.**

An RPC/indexer miss or insufficient query window produces an unknown outcome, not proof that another mint consumed the nonce. Retrying reconciliation or holding the mint for operator verification is safer; otherwise a valid mint may be closed and a replacement issuance started, creating an accounting error.

rouzwelt (changes requested):

SPEC.md
JuaniRios: Addressed in 7676425bee500867a8f00ac4e70e53d6e7a19957. issuer_request_id is now the single canonical correlation key, with explicit resolution, uniqueness/idempotency, and signed-fact mismatch rules; tokenization_request_id is explicitly not accepted for authorization correlation.

SPEC.md
JuaniRios: Addressed in 7676425bee500867a8f00ac4e70e53d6e7a19957. CloseMint now requires positive proof that every unresolved persisted transaction can no longer land—a confirmed revert or a confirmed wallet nonce advance—before closure is allowed.

SPEC.md
JuaniRios: Addressed in 7676425bee500867a8f00ac4e70e53d6e7a19957. The spec now separates invalid authorizations that arrive and are synchronously rejected from missing authorizations where no call arrives; the latter waits before submission, never falls back to vault-direct, and remains visible to stuck-mint queries.

SPEC.md
JuaniRios: Addressed in 7676425bee500867a8f00ac4e70e53d6e7a19957. RAI-1220 now explicitly blocks decoder implementation until the merged ABI confirms the exclusive range endpoint and resume-pointer semantics, with both encoded in a real Burned-log decoder test.

SPEC.md
JuaniRios: Addressed in 7676425bee500867a8f00ac4e70e53d6e7a19957. A consumed nonce with no matching Minted log is now classified as NonceReplayUnresolved and retried for reconciliation; NonceConsumedByOtherMint is reserved for an affirmative mismatching log.

SPEC.md
```diff
@@ -341,6 +453,44 @@ predecessor was `Minting`. It emits `ExistingMintRecovered`, transitions to
 `CallbackPending` boundary before delivering the callback, so receipt polling
 and Alpaca requests do not hold the wallet transaction lock.
 
+**Orchestrator mode** introduces no new commands: `Deposit`, `SubmitMint`,
+`ConfirmMint`, and `Recover` are reused unchanged, with the handler branching on
+`VaultMode` internally to submit `orchestrator.mint()` calldata instead of the
```
rouzwelt: Correct the command-count statement.

This section says orchestrator mode introduces no new commands, but AuthorizeMint is a new command in this specification. Limit the statement to the reused submission and recovery commands, or include AuthorizeMint.

SPEC.md
```diff
@@ -895,6 +1166,815 @@ action) resolve by `{underlying}:{network}` and take a required
...
+  orchestrator mint or burn, the bot checks `vaultLogicIsExpected()`. If it
+  returns `false`, the bot does not submit — there is no transaction to record,
+  so no `MintingFailed`/`BurningFailed` event is produced, the `attempts`
+  counter and its bounded 1m/10m/30m/1h exhaustion schedule do not advance, and
+  the bot logs at WARN. Recovery is deferred the same way it already is for each
+  aggregate: for mint, the existing background scheduled-recovery task re-checks
+  health before its next attempt; for burn (which has no automatic retry loop —
+  see "InsufficientReceipts" above), the next startup or manual admin recovery
+  attempt re-checks health before resubmitting. No aggregate state changes;
+  "orchestrator halted" is a global condition, not a domain fact about any one
```
rouzwelt: Scope the health gate per ChainRuntime.

This section defines VaultLogicMismatch as a global halt. The later multichain design creates separate ChainRuntime and VaultService instances per network. A mismatch on one chain must not halt unrelated chains unless that is intentional. Scope health, retry suppression, and recovery by (network, orchestrator).

SPEC.md
```diff
@@ -431,39 +599,49 @@ on-chain transfer through calling Alpaca to burning tokens.
...
-  plan persisted by the latest `BurningFailed` event exactly, with the owner
-  recovered from the transaction's own signature, and refuses a hash any other
-  redemption's history already mentions. Pre-intent states with no persisted
-  burn plan at all are **not** force-completed; ops use `CloseRedemption` after
-  off-chain reconciliation instead.
+  receipt must have succeeded and contain a real burn of the vault's shares, in
+  the shape this redemption's own persisted `burn_mode` expects
+  (`Transfer(bot_wallet -> 0x0)` for vault-direct; see "Orchestrator Migration"
+  for the orchestrator-mode proof shape) — then records the proving tx hash and
+  block number. The receipt reservation is settled (mirror reduced) just like a
```
rouzwelt: Make force-complete accounting mode-specific.

This section says every force-completed burn settles a receipt reservation and reduces the mirror. Later, orchestrator mode explicitly skips plan_burn and the entire reservation lifecycle. Define orchestrator force completion without vault-direct reservation mutation.

rouzwelt (changes requested):

0xgleb (approved):

rouzwelt (approved):

PR #221, opened 2026-07-06
replace fireblocks with turnkey

rouzwelt:
## Motivation

This PR replaces the Fireblocks signing backend with Turnkey while maintaining support for local key signing.

### **1\. Configuration & Dependencies**

- **Removed**: `FIREBLOCKS_API_USER_ID` and `fireblocks-sdk` dependency
- **Added**: Three Turnkey environment variables (`TURNKEY_API_PRIVATE_KEY`, `TURNKEY_ORG_ID`, `TURNKEY_ADDRESS`)
- **Added**: `turnkey_api_key_stamper` and `turnkey_client` (v0.4.0) dependencies
- Cleaned up ~15 unused transitive dependencies (bon, curve25519-dalek, ed25519, jsonwebtoken, pem, reqwest-middleware, etc.)

### **2\. Module Restructuring**

- **Renamed**: `fireblocks` module → `wallet`module
- **Created**: `wallet::local` and `wallet::turnkey` submodules
- **Updated**: `SignerConfig` enum now supports `Turnkey` and `Local` variants

### **3\. Core API Changes**

- **Renamed fields**: `fireblocks_tx_id` → `tx_id` throughout codebase
- **Renamed events**:
    - `FireblocksSubmitted` → `MintTxSubmitted`
    - `BurnFireblocksSubmitted` → `BurnTxSubmitted`
- **Removed**: `check_fireblocks_tx()`method and `FireblocksTxStatus` enum
- **Simplified**: Vault service interface by removing Fireblocks-specific status checking

### **4\. Recovery & Admin Logic**

- Removed Fireblocks-specific transaction status checks (pending/failed/completed)
- Removed `check_fireblocks_tx` admin endpoint
- Simplified recovery logic for both mints and redemptions
- Removed Fireblocks status enrichment from stuck transaction reports
- Updated retry logic to be backend-agnostic

### **5\. New Turnkey Implementation(`src/wallet/turnkey.rs`)**

- implementing Turnkey signing
- Uses P-256 API key authentication with secure enclave signing (50-100ms latency)
- Implements alloy's `Signer` and `TxSigner`traits
- Includes comprehensive test coverage with integration tests
- Custom retry logic for Turnkey API calls

### **6\. Documentation & Comments**

- Updated all comments to remove Fireblocks references
- Changed terminology from "Fireblocks transaction" to generic "transaction"
- Updated error messages and logging

## Checks

- [ ] made this PR as small as possible
- [ ] linked any relevant issues or PRs

## Summary by CodeRabbit

* **New Features**
  * Added Turnkey-based wallet signing and configuration (including updated secrets example).
  * Updated mint, burn, confirmation, and recovery flows to use generic `tx_id` identifiers with backward-compatible handling of legacy event field names.
  * Updated admin stuck/recovery views and logic to rely on history-derived transaction identifiers (backend-specific enrichment removed).
* **Bug Fixes**
  * Prevented infinite mint recovery retries by triggering a fresh recovery submission after terminal on-chain failures.
* **Chores / Documentation**
  * Updated the internal admin OpenAPI surface to remove backend-specific endpoints.


rouzwelt:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/221?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#225** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/225?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#226](https://github.com/ST0x-Technology/st0x.issuance/pull/226) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/226?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#227](https://github.com/ST0x-Technology/st0x.issuance/pull/227) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/227?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#224** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/224?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#223** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/223?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#123** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/123?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#118** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/118?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#230** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/230?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#229** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/229?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#228** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#221** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/221?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/221?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#211** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/211?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#204** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/204?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#234](https://github.com/ST0x-Technology/st0x.issuance/pull/234) <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/234?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=221).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/mint/event.rs
```diff
@@ -69,13 +69,13 @@ pub(crate) enum MintEvent {
         closed_at: DateTime<Utc>,
     },
 
-    /// Transaction submitted to the signing backend (Fireblocks or local).
+    /// Mint transaction submitted to the signing backend (Turnkey or local).
     /// Persists the backend transaction ID so that polling can resume after
     /// a restart without resubmitting (which would double-mint).
-    FireblocksSubmitted {
+    MintTxSubmitted {
```
0xgleb, line 78: critical: renaming this variant and its fireblocks_tx_id field breaks deserialization of every stored FireblocksSubmitted event. The enum is externally tagged, so old payloads look like {"FireblocksSubmitted": {..., "fireblocks_tx_id": ...}}, and the projection rebuild at startup fails with an unknown-variant error, a hard startup crash on any database with production history. Add #[serde(alias = "FireblocksSubmitted")] on the variant and #[serde(alias = "fireblocks_tx_id")] on tx_id, or register an upcaster.

src/redemption/event.rs
```diff
@@ -168,21 +168,21 @@ pub(crate) enum RedemptionEvent {
     },
     /// Burn transaction submitted to the signing backend.
     /// Persists the backend transaction ID so polling can resume after a restart.
-    BurnFireblocksSubmitted {
+    BurnTxSubmitted {
```
0xgleb, line 171: critical: same backward-compat break as the mint side. Old events are stored as BurnFireblocksSubmitted with a fireblocks_tx_id field and will no longer deserialize, so any redemption that reached burn submission fails aggregate replay. Needs #[serde(alias = "BurnFireblocksSubmitted")] on the variant and #[serde(alias = "fireblocks_tx_id")] on tx_id.

src/redemption/event.rs
```diff
@@ -168,21 +168,21 @@ pub(crate) enum RedemptionEvent {
...
 
-    /// Existing on-chain burn discovered during recovery via Fireblocks tx lookup.
+    /// Existing on-chain burn discovered during recovery via tx lookup.
     /// Mirrors mint's `ExistingMintRecovered` — the burn already landed on-chain
-    /// but the bot failed to record it (e.g., Fireblocks polling timeout).
+    /// but the bot failed to record it (e.g. polling timeout).
     ExistingBurnRecovered {
         issuer_request_id: IssuerRedemptionRequestId,
-        fireblocks_tx_id: String,
+        tx_id: String,
```
0xgleb: critical: tx_id here is required and has no alias, but old ExistingBurnRecovered events carry the key fireblocks_tx_id. Serde ignores the unknown key and then errors on the missing tx_id, so every redemption that went through this recovery path becomes unreadable. One line fixes it: #[serde(alias = "fireblocks_tx_id")].

src/redemption/event.rs
```diff
@@ -143,10 +143,10 @@ pub(crate) enum RedemptionEvent {
         issuer_request_id: IssuerRedemptionRequestId,
         error: String,
         failed_at: DateTime<Utc>,
-        /// Fireblocks transaction ID, if the burn was submitted via Fireblocks.
-        /// Absent for non-Fireblocks backends or pre-enrichment events.
+        /// Transaction ID
+        /// Absent for pre-enrichment events.
         #[serde(default)]
-        fireblocks_tx_id: Option<String>,
+        tx_id: Option<String>,
```
0xgleb: should fix: unlike the hard breaks on the renamed variants, this one fails silently. Because of #[serde(default)], old BurningFailed events with fireblocks_tx_id: Some(...) deserialize with tx_id = None, quietly losing the stored transaction id, and /admin/stuck then reports those redemptions as waiting for burn submission instead of confirmation. Same one-line fix: #[serde(alias = "fireblocks_tx_id")], plus a backwards-compat test that feeds the old key.

src/mint/view.rs
```diff
@@ -79,7 +79,7 @@ pub(crate) enum MintView {
         journal_confirmed_at: DateTime<Utc>,
         minting_started_at: DateTime<Utc>,
     },
-    FireblocksSubmitted {
+    MintTxSubmitted {
```
0xgleb, line 83: should fix: the view enum has the same rename problem. Existing mint_view rows were persisted with the FireblocksSubmitted variant name and stop deserializing, which makes mints stuck in that state invisible to recovery and admin queries even after the event-level aliases land. Either alias the old variant name here too or rebuild the view as part of the deploy.

src/vault/service.rs
```diff
@@ -149,29 +150,29 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
-        debug!(target: "vault", tx_hash = %fireblocks_tx_id,
+        debug!(target: "vault", tx_hash = %tx_id,
             "Mint cache empty, recovering from chain"
         );
 
         let tx_hash: B256 =
-            fireblocks_tx_id.parse().map_err(|_| VaultError::InvalidReceipt)?;
+            tx_id.parse().map_err(|_| VaultError::InvalidReceipt)?;
 
         let receipt = self
```
0xgleb: should fix: confirm_mint can't distinguish a pending tx from a failed one. If get_transaction_receipt returns nothing because the tx is still in the mempool (congestion, slow inclusion), that maps to a failure, the aggregate records MintingFailed, and recovery resubmits while the original can still mine, a double-mint window. The old flow kept a dedicated pending status with a multi-hour poll budget before concluding failure. Worth checking get_transaction_by_hash before treating a known tx as dead, and only resubmitting once it's definitively gone.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+    where
+        Request: Serialize + Sync,
+        Response: serde::de::DeserializeOwned,
+    {
+        let url = format!("{}{}", self.base_url, path);
+        let post_body = serde_json::to_string(request)?;
+        let StampHeader { name, value } =
+            self.api_key.stamp(post_body.as_bytes())?;
+        let response = self
+            .http
```
0xgleb, line 385: should fix: the signing requests never set a Content-Type: application/json header. Turnkey's API takes JSON bodies, and strict gateways reject or misroute bodies without the header, so this can fail against the real API even though the mock server accepts it.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+            "Turnkey API response body received"
+        );
+
+        // Error mapping below reproduces upstream `turnkey_client`'s handling
+        // (including its String-valued error variants); it is not a new
+        // stringly-typed-error introduction. This client is a deliberate fork
+        // of `turnkey_client` to insert the response-body trace above.
+        let content_type = content_type
+            .ok_or(TurnkeyClientError::MissingContentTypeHeader)?
+            .to_str()
```
0xgleb, line 427: should fix: process_request validates the response Content-Type before checking the HTTP status, so a 401 or 429 that comes back as text/html surfaces as MissingContentTypeHeader instead of the actual auth or rate-limit error. Check status first and include the body in the error, then validate the content type.

src/admin.rs
```diff
@@ -139,7 +131,9 @@ struct AlpacaCalledData {
 
 /// Data extracted from a BurningFailed event in the event history.
 struct BurningFailedData {
-    fireblocks_tx_id: Option<String>,
+    #[allow(dead_code)]
```
0xgleb: should fix: these #[allow(dead_code)] attributes are papering over a dead pipeline. BurningFailedData and PostAlpacaRecoveryInput.burning_failed are still populated by load_reprocess_context and accepted in the request body, but recover_post_alpaca destructures them away with .., so the API advertises a field that does nothing. Cleaner to remove BurningFailedData and the burning_failed field entirely.

src/redemption/burn_manager.rs
```diff
@@ -516,27 +478,23 @@ impl BurnManager {
         Ok(())
     }
 
-    /// Recovers a BurnFailed redemption that has a previously submitted Fireblocks
+    /// Recovers a BurnFailed redemption that has a previously submitted
     /// transaction. Tries to confirm the existing transaction rather than resubmitting.
     async fn recover_burn_failed_with_existing_tx(
```
0xgleb, line 485: should fix: recover_burn_failed_with_existing_tx has no test coverage left. The four tests that exercised the prior-tx path were deleted with the Fireblocks variants, and the surviving recover_burn_failed tests all seed tx_id: None. The two branches here (confirm succeeds and settles vs reverted, release reservation and mark failed) have materially different custody outcomes, so both deserve a test before this merges.

src/redemption/burn_manager.rs
```diff
@@ -1347,33 +1302,20 @@ const fn aggregate_state_name(aggregate: &Redemption) -> &'static str {
     }
 }
 
-/// Extracts the Fireblocks transaction ID from a `VaultError`, if the error
-/// originated from a Fireblocks error that carries a transaction ID.
-pub(crate) fn extract_fireblocks_tx_id(error: &VaultError) -> Option<String> {
+/// Extracts the transaction hash from a `VaultError`
+pub(crate) fn extract_tx_hash(error: &VaultError) -> Option<String> {
```
0xgleb: should fix: extract_tx_hash demotes a typed B256 to Option<String> for the error field. AGENTS.md calls this out explicitly, error types must store typed values, not string renderings. Keep the field as Option<B256> and let the Display impl do the formatting.

src/mint/recovery.rs
```diff
@@ -54,8 +54,6 @@ pub(crate) enum DriveOutcome {
     Done,
     /// Paused until the next automatic retry window elapses.
```
0xgleb, line 55: minor: the deleted pending-poll tests were the behavioral spec for slow-to-confirm transactions and nothing replaced them. There's now no test for TxSubmitted plus a confirm_mint failure transitioning to MintingFailed, and none for MAX_SCHEDULED_RECOVERY_FAILURE_BACKOFFS exhaustion. Worth adding both to pin the new recovery behavior.

src/mint/event.rs
```diff
@@ -69,13 +69,13 @@ pub(crate) enum MintEvent {
...
         issuer_request_id: IssuerMintRequestId,
         external_tx_id: String,
```
0xgleb, line 80: minor: external_tx_id stays a bare String on the mint side while the burn side already has the BurnExternalTxId newtype, and SubmittedTx in vault/mod.rs carries external_tx_id and tx_id as two adjacent Strings the compiler can't tell apart. A MintExternalTxId newtype mirroring the burn side would remove the swap risk on the idempotency key.

src/admin.rs
```diff
@@ -1590,29 +1395,21 @@ fn redemption_history_summary_from_events(
...
-                fireblocks_tx_id: Some(fireblocks_tx_id),
-                ..
-            } => {
-                summary.fireblocks_tx_id = Some(fireblocks_tx_id);
+            RedemptionEvent::BurnTxSubmitted { tx_id, .. }
+            | RedemptionEvent::ExistingBurnRecovered { tx_id, .. }
+            | RedemptionEvent::BurningFailed { tx_id: Some(tx_id), .. } => {
+                summary.tx_id = Some(tx_id);
             }
             _ => {}
```
0xgleb, line 1543: minor: the catch-all _ => {} means a future RedemptionEvent variant silently drops out of the stuck-redemption summary. Listing the ignored variants explicitly turns a missed variant into a compile error instead of silent wrong output for operators.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+    ) -> Result<Signature, TracingTurnkeySignerError> {
+        let r_bytes = hex::decode(&response.r)?;
+        let s_bytes = hex::decode(&response.s)?;
+        let v_bytes = hex::decode(&response.v)?;
+
+        let r_arr: [u8; 32] =
+            r_bytes.try_into().map_err(|bytes: Vec<u8>| {
+                TracingTurnkeySignerError::BadComponentLength {
+                    component: SignatureComponent::R,
+                    len: bytes.len(),
```
0xgleb: minor: parse_signature assumes r and s decode to exactly 32 bytes, but the turnkey_client types only document them as String. If the API ever returns a value without leading zero padding, the try_into fails and all signing breaks. Left-padding to 32 bytes before the conversion, or capturing a real response to confirm the padding guarantee, would close this.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+    ) -> Result<Activity, TurnkeyClientError> {
+        let mut retry_count = 0;
+
+        loop {
+            let response: ActivityResponse =
+                self.process_request(request, path).await?;
+            let activity =
+                response.activity.ok_or(TurnkeyClientError::MissingActivity)?;
+
+            match activity.status {
```
0xgleb, line 335: minor: on ActivityStatus::Pending the retry loop re-POSTs the same sign_raw_payload request instead of polling the existing activity. If Turnkey doesn't dedupe re-POSTs, each retry spawns another signing activity. Is the re-POST idempotent by API contract? If not, polling the activity status endpoint would be safer. Either way a test covering PENDING then COMPLETED would pin the behavior, the current PENDING test uses zero retries so the transition is never exercised.

src/wallet/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+        assert!(
+            matches!(result, Err(SignerConfigError::NeitherConfigured)),
+            "Expected NeitherConfigured error, got {result:?}"
+        );
+    }
+
+    #[test]
+    fn both_configured_fails() {
+        let env = SignerEnv {
+            local: LocalSignerEnv { evm_private_key: Some(B256::ZERO) },
```
0xgleb, line 197: minor: these tests cover all four error paths and the local happy path but never the Turnkey happy path. Nothing asserts that a fully populated env produces SignerConfig::Turnkey with the right address wired in, and that's the production backend. One more test closes the gap.

src/wallet/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum WalletKind {
+    Local,
+    Turnkey,
+}
+
+/// Resolved signer: an `EthereumWallet` and the corresponding address.
+///
+/// Both Rutnkey and local use this path.
```
0xgleb: nit: typo, Rutnkey.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+    /// Wallet's address
+    address: Address,
+}
+
+impl std::fmt::Debug for TurnkeyWallet {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter
+            .debug_struct("TurnkeyWallet")
+            .field("address", &self.address)
+            // .field("required_confirmations", &self.required_confirmations)
```
0xgleb: nit: leftover commented-out .field for a required_confirmations field that doesn't exist on TurnkeyWallet.

src/redemption/burn_manager.rs
```diff
@@ -1440,49 +1380,21 @@ mod tests {
...
-            assert!(
-                should_release_reserved_burn(&transaction_failed(status)),
-                "terminal status {status:?} should release the reservation"
-            );
-        }
+    fn transaction_failed() -> VaultError {
+        VaultError::Reverted { tx_hash: B256::random() }
     }
 
     #[test]
```
0xgleb, line 1660: nit: should_release_on_terminal_failure and should_release_on_evm_revert are the same test twice, both feed VaultError::Reverted into a matches! that ignores the hash. Keep one, or point the other at a different variant and assert false to widen coverage.

src/admin.rs
```diff
@@ -412,14 +402,15 @@ struct PostAlpacaRecoveryInput {
...
+    #[allow(dead_code)]
     burning_failed: Option<BurningFailedData>,
     burn_retry_external_tx_id: Option<BurnExternalTxId>,
 }
 
 async fn recover_post_alpaca(
     store: &Store<Redemption>,
     alpaca_service: &Arc<dyn AlpacaService>,
-    vault_service: &Arc<dyn VaultService>,
+    _vault_service: &Arc<dyn VaultService>,
```
0xgleb: nit: _vault_service is dead now that inspect_prior_fireblocks_burn is gone, and the call site still builds and passes it. Drop the parameter.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+        assert!(matches!(
+            error,
+            TurnkeyClientError::UnexpectedMimeType(mime) if mime == "text/html"
+        ));
+    }
+
+    #[tokio::test]
+    async fn new_constructs_wallet_with_correct_address() {
+        let anvil = Anvil::new().spawn();
+        let server = MockServer::start();
```
0xgleb, line 1106: nit: new_constructs_wallet_with_correct_address only asserts that the constructor stores the address it was given, it would pass with the entire signing implementation deleted. Either drop it or make it drive a transaction against local Anvil like the failure-path test.

0xgleb (changes requested):

src/mint/event.rs
```diff
@@ -69,13 +69,13 @@ pub(crate) enum MintEvent {
         closed_at: DateTime<Utc>,
     },
 
-    /// Transaction submitted to the signing backend (Fireblocks or local).
+    /// Mint transaction submitted to the signing backend (Turnkey or local).
```
JuaniRios, line 74: critical: this renames the persisted variant `FireblocksSubmitted` to `MintTxSubmitted` and the field `fireblocks_tx_id` to `tx_id` with no serde alias or upcaster. `MintEvent` is externally-tagged, so existing rows are stored as `{"FireblocksSubmitted": {"fireblocks_tx_id": ...}}` and won't deserialize into the new enum, which breaks aggregate replay, recovery, and stuck-listing for any mint that reached submission. We know prod has these rows (the MSTR/COIN and RAI-912 incidents were exactly these events). Add an upcaster or `#[serde(alias)]` plus a test that replays the old raw JSON, same as the `TokensBurnedDataWire` pattern already in `redemption/event.rs`.

src/redemption/event.rs
```diff
@@ -143,10 +143,10 @@ pub(crate) enum RedemptionEvent {
         issuer_request_id: IssuerRedemptionRequestId,
         error: String,
         failed_at: DateTime<Utc>,
-        /// Fireblocks transaction ID, if the burn was submitted via Fireblocks.
-        /// Absent for non-Fireblocks backends or pre-enrichment events.
+        /// Transaction ID
```
JuaniRios: critical: same rename problem here (`BurnFireblocksSubmitted` to `BurnTxSubmitted`, `fireblocks_tx_id` to `tx_id`) with no upcaster, but `BurningFailed` is worse. `tx_id` kept `#[serde(default)]`, so an old event carrying `fireblocks_tx_id` deserializes fine but silently lands as `tx_id: None` instead of erroring, which is exactly the silent-defaulting the financial-integrity rule forbids. Recovery then thinks no burn was ever submitted and can fire a second on-chain burn if the original lands. Need aliases/upcasters for all three (`BurningFailed`, `BurnTxSubmitted`, `ExistingBurnRecovered`), plus a test seeding the old key populated (not absent, the existing test only covers absent).

src/redemption/burn_manager.rs
```diff
@@ -576,7 +534,7 @@ impl BurnManager {
             }
             Err(err) => {
                 warn!(target: "redemption", issuer_request_id = %issuer_request_id,
-                    %fireblocks_tx_id,
+                    %tx_id,
```
JuaniRios, line 567: critical: `recover_burn_failed_with_existing_tx` treats any `confirm_burn` error as a dead tx and marks the redemption Failed without keeping `tx_id`. But `confirm_burn` now folds in a 30s pending-receipt timeout and RPC errors, and neither proves the tx failed or consumed no receipts. A slow-but-valid burn gets downgraded to Failed with the hash gone, so manual recovery can submit a replacement while the original is still pending, and both mine. Split terminal failures from pending/unknown: keep it in `BurnFailed` with `tx_id` and keep polling, only drop the hash on a confirmed revert.

src/vault/service.rs
```diff
@@ -149,29 +150,29 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
 
         Ok(SubmittedTx {
             external_tx_id: external_tx_id.unwrap_or_else(|| {
-                format!("local-mint-{}", receipt_info.issuer_request_id)
+                format!("mint-{}", receipt_info.issuer_request_id)
```
JuaniRios, line 131: critical: with the direct alloy path, mint/burn broadcasts and waits for a receipt before any `MintTxSubmitted`/`BurnTxSubmitted` event gets persisted. A crash in that window leaves no durable tx hash, nonce, or idempotency key, which is the exact case Fireblocks' `externalTxId` used to dedupe. On restart recovery sees no stored tx and no mined receipt yet, then submits a second one. This is the crash-safety the base branch is named for, so it needs a durable submit/outbox step that records nonce and raw tx hash before broadcast, then rebroadcasts/polls that same tx on recovery. Is there a lighter option here or does it need the full outbox?

src/mint/mod.rs
```diff
@@ -1182,28 +1100,56 @@ impl Mint {
                     warn!(
                         target: "mint",
                         issuer_request_id = %issuer_request_id,
+                        tx_id = %known_tx.tx_id,
```
JuaniRios, line 1124: should fix: same shape as the burn recovery issue. The recovery path treats any `confirm_mint(tx_id)` error as proof the mint failed and submits a retry, but `submit_mint` already broadcast and waited for a receipt before recording the event, so a confirm error can just be a transient receipt/RPC/indexing blip. That means a mint that actually succeeded on-chain can get submitted again, creating duplicate backed tokens for one journal. Don't retry from a known tx hash unless chain inspection proves it reverted, and add a test where `confirm_mint` fails once after a successful submit.

src/admin.rs
```diff
@@ -428,8 +419,8 @@ async fn recover_post_alpaca(
         issuer_request_id,
         metadata,
         alpaca_data,
-        burning_failed,
         burn_retry_external_tx_id,
+        ..
```
JuaniRios, line 422: should fix: `recover_post_alpaca` still receives the `burning_failed` context but destructures it away with `..` and no longer checks a recorded `tx_id` before `ResumeBurn`/`execute_recovered_burn`. That drops the old guard, so an operator running `/admin/recover` on a Failed redemption with a prior submitted burn can unknowingly fire a second on-chain burn. If history has a burn `tx_id`, verify that tx first, keep returning pending if there's no receipt yet, and only allow a new burn after a definitive no-consumption result or an explicit force override.

src/admin.rs
JuaniRios: should fix: `BurningFailedData::tx_id`, `BurningFailedData::planned_burns`, and `PostAlpacaRecoveryInput::burning_failed` got `#[allow(dead_code)]` because the refactor left them unused once `inspect_prior_fireblocks_burn` was deleted. AGENTS.md is zero-tolerance on `#[allow(...)]` without sign-off, and the prescribed fix is to remove the dead code, not silence it. Worth noting these are the same fields the old on-chain-status-aware recovery used, so leaving them makes the code read as if `burning_failed` still matters when it doesn't. Just drop the fields and update the construction sites/tests.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+        api_private_key: &TurnkeyApiPrivateKey,
+        organization_id: TurnkeyOrganizationId,
+        address: Address,
+        chain_id: Option<ChainId>,
+    ) -> Result<Self, TracingTurnkeySignerError> {
+        let client = TracingTurnkeyClient::from_api_key(api_private_key)?;
+        Ok(Self::new(client, organization_id, address, chain_id))
+    }
+
+    fn parse_signature(
```
JuaniRios, line 509: should fix: `parse_signature` requires `r`/`s` to decode to exactly 32 bytes and `v` to exactly one byte, with no handling for `0x` prefixes, minimal integer encodings, or odd-length hex like `"0"`/`"1"`. Turnkey's API docs describe `r`, `s`, and `v` only as strings, not fixed-width arrays, so a valid signature could get rejected before broadcast and make all Turnkey signing fail for that response shape. Normalize first: strip optional `0x`, left-pad `r`/`s` to 32 bytes when short, reject over 32, and accept `v` as hex or decimal for 0/1/27/28.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+
+        match inner {
+            result::Inner::SignRawPayloadResult(result) => Ok(result),
+            other => Err(TurnkeyClientError::UnexpectedInnerActivityResult(
+                serde_json::to_string(&other)?,
+            )),
+        }
+    }
+
+    async fn process_activity<Request: Serialize + Sync>(
```
JuaniRios, line 315: should fix: `process_activity` branches on Completed/Pending/Failed/ConsensusNeeded/catch-all, but the tests only cover pure-Completed and Pending-exhausts-all-retries. The actual retry-then-succeed path (Pending on early polls, Completed later) isn't tested, and neither are the `Failed` to `ActivityFailed` or `ConsensusNeeded` to `ActivityRequiresApproval` mappings. This is the retry logic on the signing critical path, so an off-by-one that only succeeds when the first poll is Completed, or a swapped Failed/ConsensusNeeded branch, would slip through. Add a mock-server test for pending-then-completed plus asserts on the two error mappings.

src/mint/mod.rs
```diff
@@ -142,9 +142,9 @@ pub(crate) enum Mint {
         minting_started_at: DateTime<Utc>,
     },
     /// Transaction submitted to signing backend, awaiting on-chain confirmation.
-    /// The `fireblocks_tx_id` enables recovery: on restart, the bot resumes
+    /// The `tx_id` enables recovery: on restart, the bot resumes
     /// polling this transaction instead of resubmitting (which would double-mint).
-    FireblocksSubmitted {
+    TxSubmitted {
```
JuaniRios, line 147: minor: `Mint::TxSubmitted.tx_id` is a `String` holding the same on-chain hash that `CallbackPending`/`Completed` store as `tx_hash: B256` in this same enum (also in `MintEvent::MintTxSubmitted` and `MintCommand::ConfirmMint`). So the value gets re-typed String to B256 mid-lifecycle and the aggregate does a fallible `.parse()` instead of the type guaranteeing validity end to end. Typing these as `B256` would match the sibling fields and kill the parse step.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+            .unwrap_err();
+
+        assert!(matches!(
+            error,
+            TurnkeyClientError::UnexpectedMimeType(mime) if mime == "text/html"
+        ));
+    }
+
+    #[tokio::test]
+    async fn new_constructs_wallet_with_correct_address() {
```
JuaniRios: nit: `new_constructs_wallet_with_correct_address` builds a wallet from `Address::random()` and asserts the provider returns that same address, but Turnkey addresses aren't derived here, the config value just flows straight through. Unlike `local_signer_resolves_to_correct_address`, which does a real key to address derivation, this only checks a value passes unchanged through a constructor, which the compiler already guarantees. Fold the assertion into a signing test or drop it.

src/config.rs
```diff
@@ -91,32 +91,27 @@ impl Config {
...
-                    env,
-                    http_provider.clone(),
-                    self.chain_id,
-                    oa_schema_cache,
-                )
-                .map_err(|err| ConfigError::FireblocksVault(Box::new(err)))?;
-
-                Arc::new(service)
-            }
+        let vault_service: Arc<dyn VaultService> = {
```
JuaniRios, line 95: nit: `resolve_local_signer` and `resolve_turnkey_signer` both return `SignerResolveError`, but `create_blockchain_setup` maps them into two different variants via `.map_err(|err| ConfigError::X(Box::new(err)))`, boxing the same concrete type twice. That's the verbose map_err-with-boxing the docs push away from, and it loses the concrete type. Collapse to one `SignerResolve(#[from] SignerResolveError)` and use `?` directly.

src/vault/mod.rs
```diff
@@ -373,18 +328,17 @@ impl ReceiptInformation {
 
 /// Result of submitting a transaction to the signing backend.
 ///
-/// Returned by `submit_mint` and `submit_burn`. The `fireblocks_tx_id` is
+/// Returned by `submit_mint` and `submit_burn`. The `tx_id` is
```
JuaniRios, line 355: nit: `SubmittedTx::tx_id` (plus the `confirm_mint`/`confirm_burn(tx_id: &str)` params and the `ConfirmMint`/`ConfirmBurn`/`RecordExistingBurn` command fields) stayed `String`. That made sense when it was backend-polymorphic (Fireblocks UUID vs hash), but now `RealBlockchainService` is the only impl and every producer is a real `B256`, so `service.rs` immediately does `.parse().map_err(|_| InvalidReceipt)` and throws away the real error. These are transient (not event-sourced), so tightening to `B256` needs no upcasting. Only the persisted event payloads have to stay `String` for replay.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+            .await?;
+        let status = response.status();
+        let content_type = response.headers().get(CONTENT_TYPE).cloned();
+        // Read raw bytes and parse the success body with `serde_json::from_slice`
+        // so invalid UTF-8 fails fast, matching the fail-fast convention the
+        // Alpaca clients follow. Lossy decoding is used only for the trace line
+        // and the error-body display.
+        let bytes = response.bytes().await?;
+
+        // The successful `sign_raw_payload` body contains the ECDSA signature
```
JuaniRios: nit: the client logs the full response body at TRACE for every response, and successful signing responses carry the ECDSA components. Since the signer exposes a generic `sign_hash`, those aren't guaranteed to be public on-chain tx signatures, so it brushes against the no-secrets-in-logs rule. Probably low-risk since it's TRACE only, but worth logging just status, activity id, and sanitized error instead. Flagging as disputed, lmk if you think it's a non-issue.

JuaniRios (changes requested):

src/wallet/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+pub(crate) struct SignerEnv {
+    #[clap(flatten)]
+    local: LocalSignerEnv,
+
+    #[clap(flatten)]
+    turnkey: TurnkeyEnv,
+}
+
+/// Local EVM private key signer configuration.
+#[derive(Args, Debug, Clone)]
```
0xgleb, line 44: should fix: the Turnkey newtypes all get hand-written redacting Debug impls, but the types upstream of them derive Debug with the secrets in the clear. SignerConfig::Local(B256) prints the full raw EVM key, and SignerEnv's derived Debug prints both evm_private_key and the Turnkey api_private_key verbatim. SignerConfig is pub, re-exported, and lives in Config for the process lifetime, so any future dbg! or error-context format ships the hot signing key to the log aggregator. Give SignerConfig and the env structs manual redacting Debug impls, or wrap the key in a newtype with a [REDACTED] Debug like TurnkeyApiPrivateKey.

src/redemption/burn_manager.rs
```diff
@@ -687,7 +645,7 @@ impl BurnManager {
                         RedemptionCommand::RecordBurnFailure {
                             issuer_request_id: issuer_request_id.clone(),
                             error: err.clone(),
-                            fireblocks_tx_id: Some(tx_id.clone()),
+                            tx_id: Some(tx_id.clone()),
```
0xgleb, line 661: should fix: when recovery hits a BurnIntended with a prepared-but-never-broadcast tx (crash between prepare_burn_tx and submit), ConfirmBurn fails and this stores the precomputed hash in BurnFailed.tx_id even though nothing was ever sent. The next recovery pass then treats it as an existing tx, spends 30s polling for a receipt that can never exist, and marks the redemption Failed. When the tx was never submitted, store tx_id: None so the next pass runs a fresh burn instead of confirming a phantom one.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+            });
+        }
+
+        self.sign_hash(&tx.signature_hash()).await
+    }
+}
+
+#[async_trait]
+impl Signer for TracingTurnkeySigner {
+    async fn sign_hash(&self, hash: &B256) -> SignerResult<Signature> {
```
0xgleb, line 579: minor: after parse_signature the signature is returned without recovering the address from the prehash and comparing it to self.address. A wrong parity that slips past normalize_v (only v=00 and v=02 are tested) produces a signature that broadcasts as the wrong sender and fails with a confusing RPC error instead of a clear signer-side one. One recover_address_from_prehash check here catches that class of bug before broadcast.

src/wallet/local.rs
```diff
@@ -0,0 +1,18 @@
+use crate::wallet::{ResolvedSigner, SignerResolveError, WalletKind};
```
0xgleb, line 1: nit: imports here lead with the internal crate::wallet group before the alloy externals, with no blank line. AGENTS.md mandates externals first, blank line, then internal. The sibling wallet modules in this PR follow it.

src/vault/service.rs
rouzwelt: that is not the case for mint, that only is the case for burns, mints on the other hand already have guard for that:

handled at the application layer — the receipt backfiller runs at
startup before recovery, scans the chain for Deposit events, and populates
the receipt inventory by issuer_request_id. recover_from_existing_receipt
finds the existing deposit and emits ExistingMintRecovered, preventing a
double-mint.

src/vault/service.rs
```diff
@@ -149,29 +150,29 @@ impl<P: Provider + Clone + Send + Sync + 'static> VaultService
...
             }),
-            fireblocks_tx_id: tx_hash_str,
+            tx_id: tx_hash_str,
```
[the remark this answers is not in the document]
rouzwelt: thats not the case for mints, that only happens for burns as for mints we have this guard:

handled at the application layer — the receipt backfiller runs at
startup before recovery, scans the chain for Deposit events, and populates
the receipt inventory by issuer_request_id. recover_from_existing_receipt
finds the existing deposit and emits ExistingMintRecovered, preventing a
double-mint. submit_mint also scans for an existing deposit before sending
as defense-in-depth (covers periodic-backfill gaps during live operation).

src/wallet/turnkey.rs
rouzwelt: this is taken from liquidty repo implementation, so guess its fine?

src/wallet/mod.rs
rouzwelt: turnkey_fully_configured_produces_turnkey_config_with_correct_address already applies

src/mint/event.rs
0xgleb: Addressed with serde aliases for the legacy FireblocksSubmitted variant and fireblocks_tx_id field, preserving stored event replay.

src/redemption/event.rs
0xgleb: Addressed with serde aliases for BurnFireblocksSubmitted and fireblocks_tx_id, preserving redemption replay.

src/redemption/event.rs
0xgleb: Addressed by aliasing fireblocks_tx_id onto tx_id for ExistingBurnRecovered.

src/redemption/event.rs
0xgleb: Addressed by aliasing fireblocks_tx_id on BurningFailed and keeping a backward-compatibility deserialization regression.

src/mint/view.rs
0xgleb: Addressed by aliasing the legacy FireblocksSubmitted mint-view variant.

src/vault/service.rs
0xgleb: Addressed. Receipt timeouts map to ConfirmationPending, leave the aggregate in TxSubmitted with the same transaction ID, and recovery retries confirmation instead of submitting a replacement. Covered by timeout regressions.

src/wallet/turnkey.rs
0xgleb: Addressed. Signed Turnkey requests now send Content-Type: application/json, with an HTTP-level regression.

src/wallet/turnkey.rs
0xgleb: Addressed. HTTP status is checked before response Content-Type, so non-JSON 401/429-style responses report the real status and body. Covered by a regression.

src/admin.rs
0xgleb: Addressed by wiring the data into the recovery decision instead of deleting it. The stored failed-burn transaction is now passed to inspect_prior_burn and checked on-chain before any resume.

src/redemption/burn_manager.rs
0xgleb: Addressed with prior-transaction recovery tests for successful confirmation and settlement, transient confirmation retention, and reverted transaction release/failure.

src/redemption/burn_manager.rs
0xgleb: Addressed. Transaction identities are kept as typed TxId values backed by B256 for on-chain hashes; formatting happens only in Display.

src/mint/recovery.rs
0xgleb: Addressed with regressions for TxSubmitted confirmation failure and scheduled recovery stopping after the maximum failure backoffs.

src/mint/event.rs
0xgleb: Addressed with a transparent MintExternalTxId newtype propagated through mint state, events, services, and SubmittedTx.

src/admin.rs
0xgleb: Addressed. The admin event folds now enumerate ignored variants explicitly, so future variants cause compile-time exhaustiveness failures.

src/wallet/turnkey.rs
0xgleb: Addressed. Signature components are left-padded to 32 bytes and oversized components are rejected; both paths are tested.

src/wallet/turnkey.rs
0xgleb: Addressed. The signing activity is submitted once, then Pending is polled through get_activity by organization/activity ID. The Pending-to-Completed test asserts one submit and one poll.

src/wallet/mod.rs
0xgleb: Addressed with a fully populated Turnkey configuration test that asserts the production variant and configured address.

src/wallet/mod.rs
0xgleb: Addressed: corrected the typo.

src/wallet/turnkey.rs
0xgleb: Addressed: removed the stale commented field.

src/redemption/burn_manager.rs
0xgleb: Addressed: removed the duplicate revert-release test and retained distinct terminal versus non-definitive coverage.

src/admin.rs
0xgleb: Addressed. The unused vault-service state was removed from the stuck-list route and tracing skip list.

src/wallet/turnkey.rs
0xgleb: Addressed. The constructor-only test was removed; Turnkey coverage now exercises successful signing, signer-address verification, failure paths, and Pending polling.

src/wallet/mod.rs
0xgleb: Addressed. SignerConfig and SignerEnv now use manual Debug implementations that redact the local EVM private key and Turnkey API private key; the regression asserts neither secret appears.

src/redemption/burn_manager.rs
0xgleb: Resolved with a safer crash-recovery design. A prepared transaction is never discarded as phantom: BurnIntended retains its exact signed bytes, hash, and nonce, recovery re-broadcasts it, and ambiguous confirmation keeps the reservation and transaction recoverable.

src/wallet/turnkey.rs
0xgleb: Addressed. The parsed Turnkey signature recovers its address from the prehash and rejects any address that differs from the configured signer; covered by a wrong-signer regression.

src/wallet/local.rs
0xgleb: Addressed. External alloy imports are grouped first, followed by a blank line and the internal crate::wallet import.

src/admin.rs
```diff
@@ -139,7 +131,9 @@ struct AlpacaCalledData {
 
 /// Data extracted from a BurningFailed event in the event history.
 struct BurningFailedData {
-    fireblocks_tx_id: Option<String>,
+    #[allow(dead_code)]
+    tx_id: Option<String>,
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the schema documentation now describes transaction IDs across the actual recoverable redemption states.

src/wallet/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    MissingAddress,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum SignerResolveError {
+    #[error(transparent)]
+    Turnkey(#[from] TurnkeyError),
+    #[error("invalid EVM private key")]
+    InvalidPrivateKey(#[from] alloy::signers::k256::ecdsa::Error),
+}
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the conversion variant is named for its source key error rather than the operation.

src/redemption/event.rs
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: legacy Fireblocks redemption variants and field names are accepted by the serde compatibility layer, with replay regression coverage for populated historical IDs.

src/redemption/view.rs
```diff
@@ -94,10 +94,10 @@ pub(crate) enum RedemptionView {
         alpaca_journal_completed_at: DateTime<Utc>,
         error: String,
         failed_at: DateTime<Utc>,
-        /// Fireblocks transaction ID, if a burn was already submitted before failure.
+        /// Transaction ID, if a burn was already submitted before failure.
         /// Present when the failure occurred during confirmation (after submission).
         #[serde(default)]
-        fireblocks_tx_id: Option<String>,
+        tx_id: Option<String>,
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the redemption view accepts the historical transaction-ID field name so old rows retain the submitted transaction during recovery.

src/vault/mock.rs
```diff
@@ -372,13 +355,13 @@ impl VaultService for MockVaultService {
...
-            fireblocks_tx_id: "mock-fb-mint".to_string(),
+            tx_id: MOCK_MINT_TX_HASH.to_string(),
         })
     }
 
     async fn confirm_mint(
         &self,
-        _fireblocks_tx_id: &str,
+        _tx_id: &str,
     ) -> Result<MintResult, VaultError> {
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: mock confirmations now validate the supplied transaction ID against stored submissions, and the burn mock preserves one hash through prepare, submit, and confirm.

src/admin.rs
```diff
@@ -106,22 +104,16 @@ pub(crate) struct StuckAggregate {
...
+    /// `MintTxSubmitted` event. For redemptions, populated only on
     /// `BurnFailed` (the view carries it for that variant).
     #[serde(skip_serializing_if = "Option::is_none")]
-    fireblocks_tx_id: Option<String>,
-    /// Live status of the Fireblocks transaction (subStatus, network tx
-    /// hashes). Best-effort: omitted when `fireblocks_tx_id` is missing,
-    /// the backend is non-Fireblocks, or the lookup itself fails.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    fireblocks_status: Option<FireblocksTxStatus>,
+    tx_id: Option<String>,
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the dead-code suppressions are gone and the recovery path now consumes the persisted prior-burn context instead of carrying unused fields.

src/redemption/burn_manager.rs
```diff
@@ -1347,33 +1302,20 @@ const fn aggregate_state_name(aggregate: &Redemption) -> &'static str {
...
-            crate::fireblocks::FireblocksVaultError::TransactionFailed {
-                tx_id,
-                ..
-            }
-            | crate::fireblocks::FireblocksVaultError::MissingTxHash { tx_id },
-        ) => Some(tx_id.clone()),
+        VaultError::Reverted { tx_hash }
+        | VaultError::EventNotFound { tx_hash } => Some(tx_hash.to_string()),
         _ => None,
     }
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: transaction identity is carried as the typed `TxId`/hash representation through the recovery boundary and stringified only for display or wire compatibility.

src/admin.rs
```diff
@@ -663,57 +653,41 @@ async fn inspect_prior_fireblocks_burn(
...
+                ));
+            }
+
+            warn!(target: "redemption", issuer_request_id = %issuer_request_id,
+                %tx_id,
+                error = %e,
+                "Failed to confirm previously submitted burn"
             );
-            Err(Status::BadGateway)
+            Ok(PriorBurnDisposition::ResumeWith(burn_retry_external_tx_id))
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: only a definitive on-chain revert enters the fresh retry path; ambiguous errors retain the persisted transaction and continue checking it.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+    immutable::activity::v1::{SignRawPayloadResult, result},
+    immutable::common::v1::{HashFunction, PayloadEncoding},
+};
+use turnkey_client::{RetryConfig, TurnkeyClientError};
+
+use crate::wallet::{ResolvedSigner, SignerResolveError, WalletKind};
+
+/// Turnkey organization identifier (non-secret, lives in plaintext
+/// config).
+#[derive(Debug, Clone, Deserialize)]
```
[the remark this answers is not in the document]
0xgleb: Verified but not applied: the listed Turnkey configuration types participate in the public wallet/config surface. Reducing them to crate visibility would break that exported API; implementation-only helpers remain private.

src/wallet/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+use clap::{Args, Parser};
+
+use serde::Deserialize;
+use turnkey::{
+    TurnkeyApiPrivateKey, TurnkeyConfig, TurnkeyError, TurnkeyOrganizationId,
+};
+
+/// Wallet backend discriminant. Deserialized from a `kind` field in
+/// wallet config sections. Variants are feature-gated so unconfigured
+/// backends fail at parse time with "unknown variant".
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the unsupported feature-gating claim was removed and the Turnkey variant spelling was corrected.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
+//! Turnkey secure enclave wallet.
+//!
+//! `TurnkeyWallet` submits transactions via Turnkey's AWS Nitro secure
+//! enclaves for low-latency signing (50-100ms). Yhe only difference is
+//! the signer: `TurnkeySigner` (remote signing via Turnkey API) instead
+//! of `PrivateKeySigner` (local key).
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the module docs use the correct spelling and refer to `TracingTurnkeySigner` consistently.

src/vault/mod.rs
```diff
@@ -373,18 +328,17 @@ impl ReceiptInformation {
...
-    /// Deterministic ID used for Fireblocks idempotency (`externalTxId`).
+    /// Deterministic ID used for idempotency (`externalTxId`).
     /// Format: `{operation}-{issuer_request_id}`.
     pub(crate) external_tx_id: String,
     /// Backend-specific transaction identifier.
-    /// For Fireblocks: the Fireblocks transaction ID.
-    /// For local backends: the on-chain transaction hash.
-    pub(crate) fireblocks_tx_id: String,
+    /// For local and Turnkey backends: the on-chain transaction hash.
+    pub(crate) tx_id: String,
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: `SubmittedTx` accepts the historical field name while current producers and consumers continue using `tx_id`.

Cargo.toml
```diff
@@ -54,7 +54,9 @@ base64 = "0.22.1"
 ipnetwork = "0.21.1"
 subtle = "2.6.1"
 governor = "0.10.2"
-fireblocks-sdk = { git = "https://github.com/0xgleb/fireblocks-sdk-rs.git", branch = "fix/confirming-not-terminal" }
+turnkey_api_key_stamper = { version = "0.4.0" }
+turnkey_client = { version = "0.4.0" }
```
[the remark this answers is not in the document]
0xgleb: Verified but not applied: the suggested `0.7.x` target is stale, while the current SDK line has advanced further. No concrete defect was identified in the pinned API, and changing major SDK lines would be a separate migration rather than a safe review fix.

src/wallet/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub(crate) fn address(&self) -> Result<Address, SignerResolveError> {
+        match self {
+            Self::Turnkey(config) => Ok(config.settings.address),
+            Self::Local(key) => {
+                let signer = PrivateKeySigner::from_bytes(key)?;
+                Ok(signer.address())
+            }
+        }
+    }
+}
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: external imports are contiguous and the internal Turnkey import is separated into the repository-standard group.

src/admin.rs
```diff
@@ -617,37 +607,37 @@ async fn inspect_prior_fireblocks_burn(
...
                      reconciliation may be needed after recovery."
                 );
             }
 
             info!(target: "admin", aggregate_id = %aggregate_id,
-                fireblocks_tx_id = %fb_tx_id,
-                tx_hash = ?tx_hash,
-                "Fireblocks tx already completed on-chain, recording existing burn"
+                tx_hash = ?tx_id,
+                "Transaction already completed on-chain, recording existing burn"
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: structured recovery logs consistently record backend transaction identity under `tx_id`.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+/// (secret, lives in encrypted config).
+#[derive(Clone, Deserialize)]
+#[serde(transparent)]
+pub struct TurnkeyApiPrivateKey(String);
+
+impl TurnkeyApiPrivateKey {
+    pub const fn new(value: String) -> Self {
+        Self(value)
+    }
+}
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: organization IDs and P-256 keys now use validated fallible construction and validate during deserialization, with invalid-input tests.

src/wallet/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+use turnkey::{
+    TurnkeyApiPrivateKey, TurnkeyConfig, TurnkeyError, TurnkeyOrganizationId,
+};
+
+/// Wallet backend discriminant. Deserialized from a `kind` field in
+/// wallet config sections. Variants are feature-gated so unconfigured
+/// backends fail at parse time with "unknown variant".
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum WalletKind {
```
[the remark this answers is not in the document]
0xgleb: Verified but not applied: these types form the public `Config` surface and are re-exported for consumers. Tightening them to crate visibility makes the public configuration API fail to compile, so the requested visibility change is not valid for the current interface.

src/vault/service.rs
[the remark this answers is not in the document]
0xgleb: Addressed across ff3ab7c and 577ac7d7: the local path persists the signed transaction intent before broadcast and recovery rebroadcasts or checks that same transaction, closing the crash window without claiming backend deduplication.

src/wallet/turnkey.rs
```diff
@@ -0,0 +1,1104 @@
...
+
+impl std::fmt::Debug for TurnkeyWallet {
+    fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        formatter
+            .debug_struct("TurnkeyWallet")
+            .field("address", &self.address)
+            // .field("required_confirmations", &self.required_confirmations)
+            .finish_non_exhaustive()
+    }
+}
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the leftover commented field was removed from the redacted debug implementation.

src/vault/service.rs
```diff
@@ -241,51 +242,48 @@ impl VaultService for RealBlockchainService {
...
+                target: "vault",
+                precomputeded_hash = %sendable_tx.hash,
+                actual_hash = %actual,
+                "submitted tx hash mismatch between precomputed and broadcast"
+            );
+            return Err(VaultError::TxHashMismatch {
+                actual,
+                precomputed: sendable_tx.hash,
             });
         }
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the structured field is `precomputed_hash`, and the nearby wording was cleaned up.

src/vault/mod.rs
```diff
@@ -128,18 +110,19 @@ pub(crate) trait VaultService: Send + Sync {
...
-    ) -> Result<Option<SendableTxWithHash>, VaultError> {
-        Ok(None)
-    }
+    ) -> Result<SendableTxWithHash, VaultError>;
+
+    /// Prepares asigned raw tx that is sendable via eth_sendRawTransaction
+    async fn check_tx(
+        &self,
+        _tx_hash: &str,
+    ) -> Result<TransactionReceipt, VaultError>;
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: both trait method docs now describe their real behavior and the signing typo is corrected.

src/wallet/mod.rs
```diff
@@ -0,0 +1,266 @@
...
+
+use alloy::network::EthereumWallet;
+use alloy::primitives::{Address, B256};
+use alloy::signers::local::PrivateKeySigner;
+use clap::{Args, Parser};
+
+use serde::Deserialize;
+use turnkey::{
+    TurnkeyApiPrivateKey, TurnkeyConfig, TurnkeyError, TurnkeyOrganizationId,
+};
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the Turnkey address documentation now says `env/cli arg`.

src/admin.rs
```diff
@@ -617,37 +607,37 @@ async fn inspect_prior_fireblocks_burn(
...
                         issuer_request_id: issuer_request_id.clone(),
-                        fireblocks_tx_id: fb_tx_id.clone(),
-                        tx_hash,
+                        tx_id: tx_id.clone(),
+                        tx_hash: receipt.transaction_hash(),
                         planned_burns: bf_data.planned_burns.clone(),
-                        block_number,
+                        block_number: receipt
+                            .block_number()
+                            .unwrap_or_default(),
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: a receipt without a block number now returns an explicit error and cannot persist block zero; a regression test covers the missing-field case.

src/vault/service.rs
```diff
@@ -1115,7 +1121,7 @@ mod tests {
         );
 
         let asserter = Asserter::new();
-        asserter.push_success(&0u64);
+        asserter.push_success(&555u64);
         asserter.push_success(&vec![log]);
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the chunking regression now queues the empty ranges before the later matching log, so it verifies the 50-block scan boundaries.

src/vault/mod.rs
```diff
@@ -445,13 +404,92 @@ pub(crate) enum VaultError {
...
+    }
+}
+
+impl<'de> Deserialize<'de> for TxId {
+    fn deserialize<D: serde::Deserializer<'de>>(
+        deserializer: D,
+    ) -> Result<Self, D::Error> {
+        let s = String::deserialize(deserializer)?;
+        s.parse().map_err(serde::de::Error::custom)
+    }
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: the `TxId` wire representation preserves an ambiguous hash-shaped legacy ID with an explicit discriminator, while retaining compatibility with existing textual payloads.

src/wallet/local.rs
```diff
@@ -0,0 +1,18 @@
...
+use alloy::network::EthereumWallet;
+use alloy::primitives::B256;
+use alloy::signers::Signer;
+use alloy::signers::local::PrivateKeySigner;
```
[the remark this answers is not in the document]
0xgleb: Addressed in 577ac7d7: imports are grouped at module scope with external and internal groups in repository order.

src/wallet/turnkey.rs
0xgleb: Addressed in a13071ef: signature parsing now strips optional prefixes, left-pads odd/minimal `r` and `s`, accepts decimal recovery IDs 0/1/27/28, and retains rejection tests for oversized or invalid components.

src/redemption/event.rs
0xgleb: Addressed in a13071ef: all affected redemption variants preserve historical transaction IDs through aliases/upcasting, including populated legacy payload tests.

src/mint/mod.rs
0xgleb: Addressed across ff3ab7c and a13071ef: known mint transaction hashes stay on the confirmation path after transient errors; a new transaction is only prepared after a definitive terminal outcome.

src/wallet/turnkey.rs
0xgleb: Addressed in a13071ef: successful responses log only request metadata/status; response bodies are limited to sanitized error diagnostics and signing components are not emitted on success.

src/admin.rs
0xgleb: Addressed in a13071ef: admin recovery consumes the persisted prior-burn transaction, checks it first, and does not resume with a replacement on an ambiguous result.

src/mint/mod.rs
0xgleb: Addressed in a13071ef: mint transaction identity is carried by the typed `TxId`/hash representation rather than reparsed from an unconstrained string.

src/config.rs
0xgleb: Addressed in a13071ef: configuration now has one typed `SignerResolve` conversion and uses `?` directly for both signer backends.

src/redemption/burn_manager.rs
0xgleb: Addressed across ff3ab7c and a13071ef: ambiguous confirmation errors retain the submitted burn and reservation for later polling; only a definitive revert releases it and enters terminal failure.

src/admin.rs
0xgleb: Addressed in a13071ef: the lint suppressions were removed and the prior-burn recovery data is now used by the recovery decision instead of remaining dead.

src/vault/mod.rs
0xgleb: Addressed in a13071ef: transient submission and confirmation APIs use typed transaction identity end to end, while persisted compatibility payloads retain their historical wire handling.

src/vault/service.rs
0xgleb: Addressed in ff3ab7c: the signed transaction intent and hash are persisted before broadcast, and recovery rebroadcasts or inspects that same transaction rather than constructing a replacement.

src/wallet/turnkey.rs
0xgleb: Addressed in a13071ef: tests cover pending-then-completed polling plus the failed and consensus-required error mappings.

src/mint/event.rs
0xgleb: Addressed in a13071ef: legacy mint variant and field names replay through the compatibility layer, with raw historical JSON regression coverage.

src/wallet/turnkey.rs
0xgleb: Addressed in a13071ef: the constructor-only pass-through test was removed; address correctness is exercised through signing/recovery behavior instead.

0xgleb (approved):
Re-reviewed all submitted and previously pending feedback on the final stack. The remaining Turnkey signature normalization gap is fixed in a13071ef; crash-safe recovery, historical replay compatibility, typed transaction identity, admin recovery, and config validation findings are covered by regression tests. The two non-applied suggestions have inline rationale.

0xgleb:
Follow-up for the shared Turnkey implementation: the configuration-boundary validation was mirrored to st0x.liquidity in ST0x-Technology/st0x.liquidity#1034. That PR validates organization IDs and P-256 API keys during construction/deserialization, preserves the public API through compatibility shims, and is green across backend, dashboard, hooks, and the full test matrix.

0xgleb (approved):

Cargo.toml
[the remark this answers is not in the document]
0xgleb: Correction: my prior rejection of this finding was unsupported. The 0.4.0 pins had been copied from the liquidity integration without a documented compatibility constraint. I updated turnkey_client and turnkey_api_key_stamper together to 0.7.0 in 18c26e50, refreshed Cargo.lock, adapted SignRawPayloadRequest to the new generate_app_proofs field using the SDK-default None behavior, and added an HTTP-boundary regression assertion that app proofs remain disabled unless explicitly opted in. Workspace check, strict Clippy, all tests, and nix flake check pass.

0xgleb:
Completed the remaining outside-diff review feedback in the latest update: Turnkey client and stamper are both pinned to 0.7.0; activity requests serialize generateAppProofs as null and have an HTTP-boundary regression test; MintTxSubmitted now keeps MintExternalTxId and TxId domain types in the read model; duplicated MintSubmissionInput construction is centralized in the recovery path; and the stale event/helper documentation and one-letter burn binding were corrected. The earlier dependency-thread reply has also been corrected to state 0.7.0 explicitly.

0xgleb:
One final outside-diff compatibility follow-up is now addressed: the legacy mint replay test uses the exact historical externally tagged FireblocksSubmitted JSON (including fireblocks_tx_id) rather than transforming a newly serialized event. I checked the pre-rename event definition as well; external_tx_id was a required field in every historical FireblocksSubmitted payload. The separate BurnIntended sendable_tx:null concern does not require a change because the current field is Option<SendableTxWithHash>, for which JSON null already deserializes to None.

0xgleb (approved):
Re-reviewed the current head after the 0.7.0 upgrade and final legacy raw-JSON replay test. All submitted and formerly pending feedback has now been either implemented or answered with a concrete invalid/superseded rationale; full check, strict clippy, formatting, workspace tests, and flake checks pass.

0xgleb (approved):
Re-reviewed the final restacked head. All submitted and pending review feedback is addressed; the Turnkey dependencies are on 0.7.0 and the crash-safety fixes from #211 are preserved.

0xgleb (approved):
Re-reviewed the final head after the CI action-pin update. All review feedback remains addressed; no unresolved or pending review threads remain.

JuaniRios (approved):

PR #278, opened 2026-07-20
fix(redemption): force-complete a landed burn with no persisted intent (RAI-1490)

JuaniRios:
https://linear.app/makeitrain/issue/RAI-1490

## Motivation

- A redemption whose burn landed on-chain but was never recorded cannot be recovered today. No admin tool accepts that state.
- It happens when the startup recovery timeout cancels the burn task after the burn lands, but before any burn event is written. The aggregate then sits in bare `Burning`. We hit it in prod on MSTR `0xcd3731…de359` (tx `0xad55f30f…`).
- Every tool rejects that state. `/admin/recover` returns 422, it needs `Failed`. A restart skips it forever, the on-chain balance is 0. `/admin/force-complete` returns 422 `PersistedBurnHashUnavailable`, it needs a persisted tx. `/admin/close` sets `Closed` instead of `Completed`, and strands the receipt reservation.
- The 30 second timeout has orphaned in-flight burns 3 times now (06-27, 07-17, and 07-20).

## Solution

- `force-complete` now accepts a redemption that never persisted a burn tx. It skips the persisted-identity checks in that case.
- Instead it binds the operator-supplied proving tx to the redemption **by amount**. The tx must burn exactly the shares owed (`alpaca_quantity`). It must return exactly the dust (`dust_quantity`). If not, the endpoint returns 422.
- `verify_burn_tx` already proves 3 things: the tx succeeded, the bot wallet signed it, and it burned the right vault's shares. The amount check replaces the persisted-hash anchor. The endpoint takes no new input, it already accepts `burn_tx_hash`.
- New errors `BurnManagerError::{ForceCompleteBurnAmountMismatch, ForceCompleteDustMismatch}` map to 422.
- **Wider than the fix:** `BurnManagerError::Redemption(_)` now maps to 422 too. It used to fall through to 500. That covers every redemption error, not just the new ones. Shout if you want it narrowed.
- `handle_force_complete_burn` skips the acknowledgement guard when there is no persisted tx. It rejects an acknowledgement supplied in that case, because there is nothing to acknowledge.
- `RECOVERY_TIMEOUT` goes from 30 seconds to 2 minutes (`src/lib.rs`).
- `SPEC.md` records the new `Burning -> Completed` admin transition and the by-amount rule.

Safety: the persisted-tx path does not change. The new path only runs when no tx was persisted. It needs an exact-amount on-chain burn, signed by the bot wallet, against the correct vault. Same binding guarantee, different anchor.

## Tests

- Aggregate: bare `Burning` and legacy `BurnSubmitted` with no persisted tx now record `BurnForceCompleted`. A supplied acknowledgement is rejected.
- Service: a matching amount moves the redemption to `Completed` and settles the reservation. A wrong amount returns `ForceCompleteBurnAmountMismatch`, and the state stays `Burning`.
- `map_burn_manager_error` maps the new error to 422.
- No e2e test. The path needs an orphaned aggregate that the public API cannot produce.

Follow-up, not in this PR: find the root cause of the recovery timeout, and teach the `recover-issuance` runbook to sum per-receipt burns (this one was a 4-receipt multicall).

## Checks

- [ ] made this PR as small as possible

https://claude.ai/code/session_01PnV5AJumthRk57FVpevrW6

## Summary by CodeRabbit

- **New Features**
  - Added force completion for redemptions stuck in the Burning state, including cases where the burn was completed on-chain but not recorded.
  - Supports legacy redemptions without a persisted burn transaction, provided the burned and returned amounts match exactly.

- **Bug Fixes**
  - Invalid burn or dust amounts now return a clear validation error.
  - Extended recovery time to better handle slow startup recovery processes.

### DIFF
--- a/src/redemption/burn_manager.rs
+++ b/src/redemption/burn_manager.rs
@@ -294,25 +294,65 @@ impl BurnManager {
                 }
             })?;
 
-        let persisted_burn_tx = redemption.persisted_burn_tx()?;
-        persisted_burn_tx.validate_for_owner(self.bot_wallet)?;
-        let acknowledged_unresolved_burn_tx_hash = redemption
-            .validate_force_complete_burn_hash(
+        let (underlying, network, alpaca_quantity, dust_quantity) =
+            match &redemption {
+                Redemption::Burning {
+                    metadata,
+                    alpaca_quantity,
+                    dust_quantity,
+                    ..
+                }
+                | Redemption::BurnIntended {
+                    metadata,
+                    alpaca_quantity,
+                    dust_quantity,
+                    ..
+                }
+                | Redemption::BurnSubmitted {
+                    metadata,
+                    alpaca_quantity,
+                    dust_quantity,
+                    ..
+                } => (
+                    metadata.underlying.clone(),
+                    metadata.network,
+                    alpaca_quantity.clone(),
+                    dust_quantity.clone(),
+                ),
+                other => {
+                    return Err(BurnManagerError::InvalidAggregateState {
+                        current_state: aggregate_state_name(other).to_string(),
+                    });
+                }
+            };
+
+        // A redemption that reached `BurnIntended` persisted its signed burn
+        // tx; that persisted identity anchors force-complete via the owner
+        // check and hash/acknowledgement guard. A redemption the recovery
+        // timeout orphaned in bare `Burning` never persisted one, so there is
+        // nothing to anchor against — the on-chain proof is bound to this
+        // redemption by amount instead (see below).
+        let persisted_burn_tx = redemption
+            .persisted_unresolved_burn_tx()
+            .filter(|sendable_tx| !sendable_tx.tx.is_empty());
+        let persisted_burn_hash = persisted_burn_tx.map(|tx| tx.hash);
+        let acknowledged_unresolved_burn_tx_hash = if let Some(sendable_tx) =
+            persisted_burn_tx
+        {
+            sendable_tx.validate_for_owner(self.bot_wallet)?;
+            redemption.validate_force_complete_burn_hash(
                 burn_tx_hash,
                 acknowledged_unresolved_burn_tx_hash,
-            )?;
-
-        let (underlying, network) = match &redemption {
-            Redemption::Burning { metadata, .. }
-            | Redemption::BurnIntended { metadata, .. }
-            | Redemption::BurnSubmitted { metadata, .. } => {
-                (metadata.underlying.clone(), metadata.network)
-            }
-            other => {
-                return Err(BurnManagerError::InvalidAggregateState {
-                    current_state: aggregate_state_name(other).to_string(),
-                });
-            }
+            )?
+        } else if let Some(provided) = acknowledged_unresolved_burn_tx_hash {
+            return Err(
+                RedemptionError::UnexpectedUnresolvedBurnAcknowledgement {
+                    provided,
+                }
+                .into(),
+            );
+        } else {
+            None
         };
 
         let vault =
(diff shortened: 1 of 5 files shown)


JuaniRios:
* **#278** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/278?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/278?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=278).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

0xgleb (commented):
agent misfire

rouzwelt (approved):

src/redemption/burn_manager.rs
```diff
@@ -324,9 +364,39 @@ impl BurnManager {
             .verify_burn_tx(vault, self.bot_wallet, burn_tx_hash)
             .await?;
 
+        // Without a persisted burn tx, the on-chain proof is the only anchor to
+        // THIS redemption. Bind it by amount: the proving tx must burn exactly
+        // the shares this redemption owes and return exactly its dust.
+        if persisted_burn_hash.is_none() {
+            let expected_burn = alpaca_quantity.to_u256_with_18_decimals()?;
+            if verification.shares_burned != expected_burn {
```
0xgleb, line 373: should fix: with no persisted burn tx this amount equality is the only thing binding the operator-supplied tx to this redemption, and amounts are not unique. `verify_burn_tx` above proves the tx succeeded against the right vault and owner, not which redemption it belongs to, and nothing anywhere in `src` dedups consumed burn tx hashes. So two redemptions on the same underlying and network, both orphaned in bare `Burning` with the same `alpaca_quantity` and `dust_quantity`, can both be force-completed with one burn tx: one on-chain burn closing two redemptions, a 1:1 backing violation.

The fix is cheap because the hash is already persisted: `BurnForceCompleted` carries `burn_tx_hash` and the `Completed` state keeps it, so a uniqueness check across redemptions is implementable today. Stronger still would be binding to the per-receipt `Withdraw` entries `verify_burn_tx` already returns, which identify the redemption rather than merely its size.

src/redemption/burn_manager.rs
```diff
@@ -324,9 +364,39 @@ impl BurnManager {
...
+                    BurnManagerError::ForceCompleteBurnAmountMismatch {
+                        expected: expected_burn,
+                        found: verification.shares_burned,
+                    },
+                );
+            }
+
+            let expected_dust = dust_quantity.to_u256_with_18_decimals()?;
+            let returned_dust = verification
+                .share_transfers
```
0xgleb, line 384: should fix: this sums every non-burn transfer from the issuer without checking its recipient. The normal burn path sends dust specifically to `metadata.wallet` (`MultiBurnParams.user`), so a different redemption's tx with the same burn and dust amounts but a different recipient can pass and terminalize this aggregate. Carry the expected wallet into this branch and require the exact transfer shape: no transfer for zero dust, otherwise one transfer of `expected_dust` to that wallet. Add wrong-recipient and non-zero-dust tests.

0xgleb (changes requested):

JuaniRios:
Closing this one, it fixes a state we can't reach anymore.

The case here was a redemption stuck in bare `Burning` with the burn already on-chain and no persisted burn tx. Only the custodian flow could produce that. Fireblocks broadcast for us, so the tx went out before we could record any tx identity. If the recovery timeout killed the task in that window, the burn landed and we had nothing written down.

That window is closed since #211. The signer now hands us the signed tx before anything is broadcast, so `persist_burn_intention` writes `BurnIntended` first and `submit_burn` broadcasts after. Two cases left:

1. Task dies before the persist. Nothing was broadcast, so nothing landed. `CloseRedemption` is the right call.
2. Task dies after the persist. State is `BurnIntended` with a persisted tx, which recovery already handles.

The old variant that got far enough to write a burn plan is covered by #299. That one binds the proving hash to the `planned_burns` in `BurningFailed`. The amount binding in this PR was only for the case with no signed tx and no plan at all, and we have no redemption in that state.

Reopen if one ever shows up, no other tool reaches it.

PR #293, opened 2026-07-27
docs: spec the staging environment

0xgleb:
## Motivation

Staging already exists and already runs Alpaca sandbox mint and redeem
validation. Its limitation is the network: it runs on production chains, so the
cash-settlement path can only be exercised with production-chain value. That
limit is what creates pressure to validate in production instead, by registering
an asset there that exists only for testing — sharing the production event
store, signer, and Alpaca account with real Authorized Participant traffic.

A test-network staging environment was previously unreachable. While the chain
registry holds a single hardcoded Base entry, the network the broker settles
cash on is reachable from the trading chain only by bridging, and no bridge path
exists between the corresponding test networks. Multichain support removes that
constraint, so the desired end state is worth agreeing before the work is
planned.

## Solution

Adds an `## Environments` section to `SPEC.md` describing the intended end state.

- Staging moves to a test network chosen so the vault chain, the liquidity trading chain, and Alpaca sandbox cash settlement coincide, removing cross-chain transport from staging.
- Enumerates what that network must supply beyond the three colocated roles: RPC, a subgraph for receipt-metadata schema resolution, signer support for its chain id, and obtainable gas.
- Records the isolation boundary as it actually is: the shared operator key decrypts both environments and the Terraform state, and staging's public surface is currently *larger* than production's because it also serves the OpenAPI schema unauthenticated.
- Specifies the missing binding — nothing today stops a staging host being configured with production endpoints — as a deploy-time gate.
- Binds promotion to a revision: the production tag must point at the exact commit that passed staging.
- Separates the standing staging environment from qualifying a *new production chain*, which still has to happen against that chain itself.
- States invariants and rejected alternatives, and marks the not-yet-true parts (trunk tracking, the smoke gate, staging-tagged telemetry) as targets rather than asserting them.

Spec only, no behaviour change. Delivery stages will be tracked in Linear rather
than in the spec.

Pairs with the `st0x.liquidity` multichain spec, which owns the settlement-topology
half of the same picture.

## Checks

By submitting this for review, I'm confirming I've done the following:

- [ ] added comprehensive test coverage for any changes in logic — n/a, spec only
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

PR #289, opened 2026-07-22
e2e tests and SPEC/README update

rouzwelt:
## Motivation

The `GET /admin/orchestrator-health` endpoint existed as a deferred item (RAI-1219) referenced throughout the spec and README, leaving operators without a concrete way to distinguish a halted orchestrator from an ordinary stuck transaction. Additionally, the spec's description of `RecordExistingBurn` and `RecordExistingBurn` (orchestrator mode) lacked detail about the `ExistingBurnProof` payload variants and their cross-checks against the persisted `burn_mode` anchor.

## Solution

The `GET /admin/orchestrator-health` endpoint is now implemented and documented. It reports, per distinct orchestrator, the `vaultLogicIsExpected()` result, and per enabled asset its resolved `vault_mode` plus — for orchestrator-mode assets — the orchestrator address and `nextBurnReceiptId`. Any on-chain read failure returns `502` rather than fabricating a health flag.

The spec and README are updated to reflect that recovery, on-chain burn verification, and the admin surface are all mode-aware: each redemption's recovery and force-completion derive from its own persisted `burn_mode` (never live config), so vault-direct and orchestrator redemptions recover side by side during incremental per-asset cutover. The previously deferred language around RAI-1219 is replaced with concrete descriptions of the implemented endpoint and its behavior.

The `RecordExistingBurn` spec notes are expanded to describe the `ExistingBurnProof::VaultDirect { burns }` and `ExistingBurnProof::Orchestrator { shares_burned, burn_range, dust_retained }` payload variants and their cross-checks against the `Failed` state's persisted `burn_mode`.

A new end-to-end test (`orchestrator_crash_recovery_confirms_in_flight_burn_failed`) covers the kill-and-restart recovery path: a redemption whose orchestrator burn landed on-chain but crashed as `BurnFailed` with the tx hash retained is confirmed by startup recovery without resubmitting, drains from `/admin/stuck`, produces exactly one `Burned` log, and the health endpoint reports the orchestrator healthy with an advanced `nextBurnReceiptId`. A `fetch_orchestrator_health` helper is introduced to ensure a broken endpoint can never silently read as healthy.

Closes [RAI-1219](https://linear.app/makeitrain/issue/RAI-1219/4-adapt-burn-verification-recovery-and-admin-endpoints)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

[...]

PR #288, opened 2026-07-22
Add `/admin/orchestrator-health` endpoint

rouzwelt:
## Motivation

During the orchestrator migration, operators need visibility into two things that are not currently surfaced: whether each orchestrator contract is healthy (i.e. `vaultLogicIsExpected()` returns `true`) and which burn path each enabled asset is using under the live configuration. Without this, distinguishing a halted orchestrator from a stuck transaction requires manual on-chain inspection.

## Solution

Adds a `GET /admin/orchestrator-health` endpoint that returns:

- A deduplicated list of orchestrator contracts with their `vaultLogicIsExpected()` health flag. A `false` value means the orchestrator is halted pending upgrade, which is visibly distinct from a stuck transaction reported by `/admin/stuck`. RPC failures surface as `502` rather than a fabricated flag.
- A per-enabled-asset list showing each asset's resolved burn path (`vault_direct` or `orchestrator`) from the live config. Orchestrator-mode rows additionally include the orchestrator address and `nextBurnReceiptId` for that asset's vault.

`vault_logic_is_expected` and `next_burn_receipt_id` are added to the `VaultService` trait with real implementations in `RealBlockchainService` that call the corresponding `IST0xOrchestratorV1` contract methods. The mock is extended with configurable return values, an error mode, and a call counter so the deduplication behaviour can be verified in tests.

The endpoint and its response types are registered in the OpenAPI schema.

Closes [RAI-1516](https://linear.app/makeitrain/issue/RAI-1516/admin-orchestrator-health-per-asset-mode-surface)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

## Summary by CodeRabbit

* **New Features**
  * Added an admin health endpoint for monitoring orchestrators and enabled asset burn paths.
  * Displays vault modes, orchestrator status, associated assets, and upcoming burn receipt IDs.
  * Supports mixed direct-vault and orchestrator configurations without duplicate health checks.
  * Documents the endpoint and response schemas in the API specification.

* **Bug Fixes**
  * Reports affected health or asset details as unavailable when RPC queries fail.
  * Correctly records transaction identifiers for stuck redemption summaries.

* **Tests**
  * Added coverage for authorization, RPC failures, halted orchestrators, mixed vault modes, and transaction ID extraction.

PR #287, opened 2026-07-22
feat: orchestrator-aware balance checks and deferral in burn recovery

rouzwelt:
## Motivation

The on-chain balance check used before attempting a burn recovery treated vault-direct and orchestrator modes identically. For vault-direct, a low balance is a reasonable signal that a broadcast may have landed without being recorded, so auto-failing or skipping is the safe choice. For orchestrator burns, this reasoning does not apply: the persist-before-broadcast invariant means nothing is ever signed before state is persisted, so a low balance cannot indicate a prior unrecorded burn — it simply means the wallet is underfunded or shares were consumed elsewhere. Auto-failing in that case would record a false "burn landed" claim.

Similarly, the required share balance for the balance check was always `burn_shares + dust_shares` regardless of mode, but orchestrator burns never move dust on-chain (it stays in the bot wallet), so the dust was incorrectly included in the threshold.

The `recover_stuck_reservations` sweep also lacked documentation clarifying why it must remain active during the incremental per-asset cutover to orchestrator mode.

## Solution

The balance check before burn recovery is now mode-aware:

- **Vault-direct**: `required_shares = burn_shares + dust_shares` (unchanged behaviour). A low balance auto-fails (`BurnFailed`) or skips (`Burning`) as before, with reservations left in place.
- **Orchestrator**: `required_shares = burn_shares` only (dust excluded). A low balance defers — the redemption stays in its current state (`BurnFailed` or `Burning`) and the next recovery pass retries once the wallet is funded. No auto-fail is issued, since no burn was ever submitted.

Documentation was added to `recover_stuck_reservations` explaining that it is inherently vault-direct-only and must remain active for any asset still using vault-direct, including during the incremental cutover period, until the last asset moves to orchestrator mode (issue 9).

Three new tests cover the orchestrator-specific behaviour:

- `recover_burn_failed_orchestrator_defers_on_insufficient_balance`: a `BurnFailed` orchestrator redemption with zero balance defers with a WARN log and does not auto-fail.
- `recover_burn_failed_orchestrator_ignores_dust_in_balance_check`: a wallet holding exactly `burn_shares` (below `burn_shares + dust`) still proceeds to completion, confirming dust is excluded from the orchestrator threshold.
- `recover_burning_orchestrator_defers_on_insufficient_balance`: a `Burning` orchestrator redemption with an underfunded wallet stays `Burning` and emits a WARN log without submitting a burn.
[...]

PR #296, opened 2026-07-27
feat: orchestrator mint preparation, confirmation, and authorization validation

rouzwelt:
## Motivation

The orchestrator contract supports a `mint()` path that is distinct from vault-direct deposits: it asserts 1:1 share minting on-chain, keeps the receipt, and forwards ERC-20 shares directly to the recipient. The service layer had no support for preparing, confirming, or recovering orchestrator mints, and there was no validation path for the recipient-produced `MintAuthV1` authorization that the orchestrator requires before it will execute a mint.

Additionally, the mint-side revert reasons (`NonceReplayed`, `BadRecipientSignature`, `RecipientCallbackRejected`, `VaultAmountMismatch`) were absent from `OrchestratorRevertReason`, meaning a reverted mint transaction could not be classified and the burn failure classifier would panic on an incomplete match.

## Solution

Four new `VaultService` methods are introduced:

- `prepare_orchestrator_mint_tx` — builds and signs the `ST0xOrchestrator.mint(token, to, amount, auth, receiptInformation)` transaction without broadcasting it. The calldata carries exactly the signed `(token, to, amount, nonce)` from the authorization; there is no `previewDeposit` and no multicall.
- `confirm_orchestrator_mint` — waits for the transaction receipt, parses the `Minted` event, and on a mined-but-reverted transaction replays it as an `eth_call` to decode the typed revert reason.
- `find_orchestrator_minted_log` — queries `Minted` logs filtered by `(token, to)` and full-matches `(to, nonce, token, amount)`. A bare `(to, nonce)` hit whose token or amount differs is explicitly not a match, since the on-chain uniqueness key is only `(to, nonce)` and a partial hit means a different mint consumed the nonce.
- `validate_mint_authorization` — checks the consumed-nonce flag first, then for a non-empty signature recovers the signer via ECDSA (EOA) or ERC-1271 `isValidSignature` (contract recipient). An empty signature skips recovery entirely, covering the future bridge/`IMintRecipient` path where the recipient authorizes via the orchestrator's `authorizeMint` callback.

The `OrchestratorRevertReason` enum gains `NonceReplayed`, `BadRecipientSignature`, `RecipientCallbackRejected`, and `VaultAmountMismatch`. The burn failure classifier's `match` is updated to explicitly route all mint-side variants to `Unclassified`, with a comment explaining they cannot arise from a burn. Three new `VaultError` variants are added: `MintAuthSignerMismatch`, `MintAuthNonceUsed`, and `MintAuthRejectedByContract`. The `IERC1271` ABI binding is wired in for the ERC-1271 contract-recipient path.

The mock service is extended with full implementations of all four methods, configurable failure injection (`with_orchestrator_mint_confirm_revert`, `with_mint_auth_failure`, `with_minted_log`, `with_orchestrator_mint_result`), call counters, and `reset()` coverage for all new state.  
  
Closes [RAI-1615](https://linear.app/makeitrain/issue/RAI-1615/orchestrator-mint-preparation-confirmation-and-authorization)

## Checks

## Summary by CodeRabbit

- **New Features**
    - Added support for orchestrator-driven minting, including transaction preparation, confirmation, minted-log recovery, and tracking mint results (nonce, amount, shares, gas, block, tx identifiers).
    - Added recipient authorization verification via signatures, including ERC-1271 contract wallet support.
- **Bug Fixes**
    - Expanded typed revert decoding and failure classification for mint scenarios (replayed nonces, invalid recipient signatures, rejected callbacks, amount mismatches) and burn-side edge cases.
- **Tests**
    - Added extensive unit-test coverage for the full mint lifecycle and log/authorization matching semantics.

### DIFF
--- a/src/redemption/mod.rs
+++ b/src/redemption/mod.rs
@@ -409,7 +409,13 @@ const fn burn_failure_classification(
             OrchestratorRevertReason::ReceiptLogicMismatch => {
                 BurnFailureClassification::ReceiptLogicMismatch
             }
-            OrchestratorRevertReason::Unknown => {
+            // Mint-side reverts can never arise from a burn transaction;
+            // if one somehow surfaces here it stays unclassified.
+            OrchestratorRevertReason::NonceReplayed { .. }
+            | OrchestratorRevertReason::BadRecipientSignature
+            | OrchestratorRevertReason::RecipientCallbackRejected { .. }
+            | OrchestratorRevertReason::VaultAmountMismatch { .. }
+            | OrchestratorRevertReason::Unknown => {
                 BurnFailureClassification::Unclassified
             }
         },
--- a/src/vault/mock.rs
+++ b/src/vault/mock.rs
@@ -16,14 +16,18 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 use tokio::sync::Notify;
 
 use super::{
-    BurnRange, BurnTxStatus, BurnVerification, MintResult, MintTxStatus,
-    MultiBurnParams, MultiBurnResult, MultiBurnResultEntry,
-    OrchestratorBurnParams, OrchestratorBurnReadiness, OrchestratorBurnResult,
-    PreparedMintTx, ReceiptInformation, SubmittedTx, VaultError, VaultService,
-    WalletNonceGuard,
+    BurnRange, BurnTxStatus, BurnVerification, MintAuthorization, MintResult,
+    MintTxStatus, MintedLogQuery, MintedLogScan, MultiBurnParams,
+    MultiBurnResult, MultiBurnResultEntry, OrchestratorBurnParams,
+    OrchestratorBurnReadiness, OrchestratorBurnResult, OrchestratorMintParams,
+    OrchestratorMintResult, PreparedMintTx, ReceiptInformation, SubmittedTx,
+    VaultError, VaultService, WalletNonceGuard,
 };
 #[cfg(test)]
-use super::{OrchestratorRevertReason, VerifiedBurn, VerifiedShareTransfer};
+use super::{
+    OrchestratorMintedLog, OrchestratorRevertReason, VerifiedBurn,
+    VerifiedShareTransfer,
+};
 use crate::redemption::BurnExternalTxId;
 use crate::vault::orchestrator::BurnProofKind;
 use crate::vault::{SendableTxWithHash, TxId};
@@ -78,6 +82,11 @@ enum MockBehavior {
     /// where signed-tx preparation fails before the tx is stored in the event.
     #[cfg(test)]
     PrepareTxFails,
+    /// Mint preparation returns a syntactically complete envelope whose
+    /// persisted hash does not match its bytes, testing the
+    /// `PreparedMintTx::validate` rejection path.
+    #[cfg(test)]
+    InvalidPreparedMint,
 }
 
 /// Configured outcome for `verify_burn_tx` in tests.
@@ -178,6 +187,45 @@ struct OrchestratorMockState {
     next_burn_receipt_id_should_error: Mutex<bool>,
     #[cfg(test)]
     vault_logic_call_count: AtomicUsize,
+    /// Cached result from `prepare_orchestrator_mint_tx` for retrieval in
+    /// `confirm_orchestrator_mint`.
+    pending_mint_result: Mutex<Option<OrchestratorMintResult>>,
+    /// When set, `confirm_orchestrator_mint` fails with
+    /// `VaultError::OrchestratorReverted` carrying this reason.
+    #[cfg(test)]
+    mint_confirm_revert: Mutex<Option<OrchestratorRevertReason>>,
+    /// Landed mint returned by `find_orchestrator_minted_log`; `None` means
+    /// no full match on-chain.
+    #[cfg(test)]
+    minted_log: Mutex<Option<OrchestratorMintedLog>>,
+    /// Optional `(orchestrator, to, token)` binding for the configured
+    /// `minted_log`: when set, a query naming a different orchestrator,
+    /// recipient, or token does NOT match — mirroring the address half of
+    /// the real lookup's four-field full-match, which the log's own fields
+    /// (`nonce`, `shares_minted`) cannot express.
+    #[cfg(test)]
+    minted_log_binding: Mutex<Option<(Address, Address, Address)>>,
+    /// When set, `validate_mint_authorization` fails with this outcome.
+    #[cfg(test)]
+    mint_auth_failure: Mutex<Option<MockMintAuthFailure>>,
+    #[cfg(test)]
+    last_mint_params: Mutex<Option<OrchestratorMintParams>>,
+    #[cfg(test)]
+    mint_preparation_call_count: AtomicUsize,
+    #[cfg(test)]
+    mint_auth_validation_call_count: AtomicUsize,
+    #[cfg(test)]
+    find_minted_log_call_count: AtomicUsize,
+}
+
+/// Configurable failure outcomes for the mock's
+/// `validate_mint_authorization`, mirroring the typed `VaultError` variants
+/// the endpoint maps to actionable responses.
+#[cfg(test)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) enum MockMintAuthFailure {
+    SignerMismatch,
+    NonceUsed,
 }
 
 /// Mock blockchain service for testing.
@@ -400,6 +448,13 @@ impl MockVaultService {
     }
 
     #[cfg(test)]
+    pub(crate) fn new_invalid_prepared_mint() -> Self {
+        let mut service = Self::new_success();
+        service.behavior = MockBehavior::InvalidPreparedMint;
+        service
+    }
+
+    #[cfg(test)]
     pub(crate) fn new_prepare_tx_failure() -> Self {
         let mut service = Self::new_success();
         service.behavior = MockBehavior::PrepareTxFails;
@@ -498,6 +553,21 @@ impl MockVaultService {
         *self.orchestrator.next_burn_receipt_id_should_error.lock().unwrap() =
             false;
         self.orchestrator.vault_logic_call_count.store(0, Ordering::Relaxed);
+        *self.orchestrator.pending_mint_result.lock().unwrap() = None;
+        *self.orchestrator.mint_confirm_revert.lock().unwrap() = None;
+        *self.orchestrator.minted_log.lock().unwrap() = None;
+        *self.orchestrator.minted_log_binding.lock().unwrap() = None;
+        *self.orchestrator.mint_auth_failure.lock().unwrap() = None;
+        *self.orchestrator.last_mint_params.lock().unwrap() = None;
+        self.orchestrator
+            .mint_preparation_call_count
+            .store(0, Ordering::Relaxed);
+        self.orchestrator
+            .mint_auth_validation_call_count
+            .store(0, Ordering::Relaxed);
+        self.orchestrator
+            .find_minted_log_call_count
+            .store(0, Ordering::Relaxed);
         *self.last_burn_proof_kind.lock().unwrap() = None;
     }
 
@@ -788,6 +858,93 @@ impl MockVaultService {
         *self.orchestrator.readiness.lock().unwrap() = Some(readiness);
     }
 
+    /// Configures `confirm_orchestrator_mint` to fail with
+    /// `VaultError::OrchestratorReverted` carrying the given typed reason.
+    #[cfg(test)]
+    pub(crate) fn with_orchestrator_mint_confirm_revert(
+        self,
+        reason: OrchestratorRevertReason,
+    ) -> Self {
+        *self.orchestrator.mint_confirm_revert.lock().unwrap() = Some(reason);
+        self
+    }
+
+    /// Overrides the result `confirm_orchestrator_mint` returns.
+    #[cfg(test)]
+    pub(crate) fn with_orchestrator_mint_result(
+        self,
+        result: OrchestratorMintResult,
+    ) -> Self {
+        *self.orchestrator.pending_mint_result.lock().unwrap() = Some(result);
+        self
+    }
+
+    /// Configures the landed mint `find_orchestrator_minted_log` can find.
+    /// Like the real implementation, the lookup only reports it when the
+    /// query full-matches — on the fields the log carries: its `nonce` and
+    /// `shares_minted` (amount) must equal the queried ones. A log configured
+    /// with a different nonce or amount exercises the
+    /// "nonce consumed by a different mint" path. Pair with
+    /// [`Self::with_minted_log_binding`] to also pin the query's address
+    /// fields (`orchestrator`, `to`, `token`), which the log itself cannot
+    /// express.
+    #[cfg(test)]
+    pub(crate) fn with_minted_log(self, log: OrchestratorMintedLog) -> Self {
+        *self.orchestrator.minted_log.lock().unwrap() = Some(log);
+        self
+    }
+
+    /// Binds the configured `minted_log` to an exact
+    /// `(orchestrator, to, token)`: a lookup passing any other address must
+    /// miss (or mismatch, for `token`), mirroring the address half of the
+    /// real scan's four-field full-match.
+    #[cfg(test)]
+    pub(crate) fn with_minted_log_binding(
+        self,
+        orchestrator: Address,
+        to: Address,
+        token: Address,
+    ) -> Self {
+        *self.orchestrator.minted_log_binding.lock().unwrap() =
+            Some((orchestrator, to, token));
+        self
+    }
+
+    /// Configures `validate_mint_authorization` to fail with the given typed
+    /// outcome.
+    #[cfg(test)]
+    pub(crate) fn with_mint_auth_failure(
+        self,
+        failure: MockMintAuthFailure,
+    ) -> Self {
+        *self.orchestrator.mint_auth_failure.lock().unwrap() = Some(failure);
+        self
+    }
+
+    #[cfg(test)]
+    pub(crate) fn get_last_orchestrator_mint_params(
+        &self,
+    ) -> Option<OrchestratorMintParams> {
+        self.orchestrator.last_mint_params.lock().unwrap().clone()
+    }
+
+    #[cfg(test)]
+    pub(crate) fn orchestrator_mint_preparation_call_count(&self) -> usize {
+        self.orchestrator.mint_preparation_call_count.load(Ordering::Relaxed)
+    }
+
+    #[cfg(test)]
+    pub(crate) fn mint_auth_validation_call_count(&self) -> usize {
+        self.orchestrator
+            .mint_auth_validation_call_count
+            .load(Ordering::Relaxed)
+    }
+
+    #[cfg(test)]
+    pub(crate) fn find_minted_log_call_count(&self) -> usize {
+        self.orchestrator.find_minted_log_call_count.load(Ordering::Relaxed)
+    }
+
     #[cfg(test)]
     pub(crate) fn last_orchestrator_burn_params(
         &self,
@@ -834,6 +991,16 @@ fn default_orchestrator_burn_result() -> OrchestratorBurnResult {
             first_receipt_id: U256::from(1u8),
             next_burn_receipt_id_after: U256::from(2u8),
         },
+        gas_used: 50000,
+        block_number: 5000,
+    }
+}
+
+fn default_orchestrator_mint_result() -> OrchestratorMintResult {
+    OrchestratorMintResult {
+        tx_hash: MOCK_MINT_TX_HASH,
+        nonce: B256::with_last_byte(1),
+        shares_minted: U256::from(100_000_000_000_000_000_000u128),
         gas_used: 50000,
         block_number: 5000,
     }
@@ -1034,7 +1201,10 @@ impl VaultService for MockVaultService {
             #[cfg(test)]
             MockBehavior::WalletLockBlocked { .. }
             | MockBehavior::SubmitRevert
-            | MockBehavior::PrepareTxFails => Err(VaultError::InvalidReceipt),
+            | MockBehavior::PrepareTxFails
+            | MockBehavior::InvalidPreparedMint => {
+                Err(VaultError::InvalidReceipt)
+            }
         }
     }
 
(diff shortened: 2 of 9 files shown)


rouzwelt:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/296?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#296** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/296?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/296?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=296).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/vault/mod.rs
```diff
@@ -262,6 +263,74 @@ pub(crate) trait VaultService: Send + Sync {
...
+        &self,
+        _orchestrator: Address,
+        _token: Address,
+        _to: Address,
+        _amount: U256,
+        _authorization: &MintAuthorization,
+    ) -> Result<(), VaultError> {
+        Err(VaultError::InvalidReceipt)
+    }
 }
```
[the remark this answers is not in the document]
rouzwelt: its intentional and will be removed in upstack PRs

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+                )
+                .await;
+            return Err(VaultError::OrchestratorReverted { tx_hash, reason });
+        }
+
+        let minted = receipt
+            .inner
+            .logs()
+            .iter()
+            .find_map(|log| {
```
0xgleb, line 1189: critical: this takes the first log in the receipt that decodes as `Minted` and lifts `nonce` and `amount` straight into the result, with no check that `log.address()` is the orchestrator, and it cannot cross-check `to`, `token` or `amount` because the method only receives `tx_id`. The mint transaction calls into the recipient at `to` before the mint completes, via ERC-1271 today and `authorizeMint` on the bridge path, so a contract at `to` emitting a log with the same topic0 and a larger amount gets picked up ahead of the orchestrator's real event, and we then record `OrchestratorTokensMinted` with shares that were never minted. `find_orchestrator_minted_log` below does this correctly, address-scoped and full-matching all four fields, so the primary success path validates strictly less than the recovery path for the same fact. Pass the expected mint facts in and share one helper between the two.

src/vault/mod.rs
```diff
@@ -262,6 +263,74 @@ pub(crate) trait VaultService: Send + Sync {
...
+    /// `MintTxIntended`) before [`VaultService::submit_mint`], which
+    /// rebroadcasts them mode-agnostically. No `previewDeposit`, no
+    /// multicall — the orchestrator asserts 1:1 on-chain and forwards shares
+    /// to the recipient. The call uses exactly the signed
+    /// `(token, to, amount, nonce)` from the authorization.
+    ///
+    /// Production callers land with the orchestrator mint aggregate arms
+    /// (mirroring how `check_orchestrator_burn_readiness` above awaited its
+    /// caller); the allow is removed then.
+    #[allow(dead_code)]
```
0xgleb, line 300: minor: this and the three sibling methods below each carry `#[allow(dead_code)]`, and a repo-wide grep at this commit shows none has a production caller, only `#[cfg(test)]` ones. AGENTS.md marks the lint policy CRITICAL and says never to add an `#[allow(...)]` without explicit permission. The justification in the comment, mirroring how `check_orchestrator_burn_readiness` awaited its caller, does not hold up: that method carries no suppression because it landed already wired into `burn_manager`.

Stack note: #300 removes all three suppressions once the methods have production callers, so the end state is clean and this is a property of the intermediate commit only. Raising it because the stated precedent is inaccurate rather than because the attribute survives.

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+            return Err(VaultError::MintAuthNonceUsed {
+                to,
+                nonce: authorization.nonce,
+            });
+        }
+
+        // The bridge/`IMintRecipient` recipient authorizes on-chain via the
+        // orchestrator's `authorizeMint` callback — there is no signature to
+        // recover (SPEC Decision 1).
+        if authorization.signature.is_empty() {
```
0xgleb, line 1358: should fix: this returns `Ok` after only the consumed-nonce check, without confirming `to` actually has contract code. For a plain EOA recipient that authorization can never be satisfied on-chain, so it passes preflight and then reverts when the mint is broadcast, burning gas and pushing the failure into revert classification instead of catching it here. Gate this branch on `to` having code, using the same `get_code_at` call the branch below already makes.

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+    async fn find_orchestrator_minted_log(
+        &self,
+        orchestrator: Address,
+        to: Address,
+        nonce: B256,
+        token: Address,
+        amount: U256,
+    ) -> Result<Option<OrchestratorMintedLog>, VaultError> {
+        let contract = IST0xOrchestratorV1::new(orchestrator, &self.provider);
+        let head = self.provider.get_block_number().await?;
```
0xgleb: should fix: this is the latest block, not a finalized or safe tag, and it becomes both the scan ceiling and the proof horizon. So a `Minted` log from a block that later reorgs out still counts as proof the mint landed, and recovery completes a mint whose shares no longer exist on chain. Query against a finality-tagged block, or require a confirmation depth before treating the log as proof.

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+                .Minted_filter()
+                .topic2(token)
+                .topic3(to)
+                .from_block(from_block)
+                .to_block(to_block)
+                .query()
+                .await?;
+
+            for (minted, log) in logs {
+                if minted.nonce == nonce
```
0xgleb: should fix: when a log matches `(to, nonce)` but the token or amount differs, this condition simply falls through and the scan continues, so the caller cannot distinguish a near-miss from nothing found. On-chain uniqueness is only `(to, nonce)`, so a near-miss means a different mint already consumed that nonce, which is a materially different situation from the mint not having landed yet. Return a distinct error for that case so recovery escalates instead of retrying.

src/vault/mod.rs
```diff
@@ -262,6 +263,74 @@ pub(crate) trait VaultService: Send + Sync {
...
+    }
+
+    /// Looks up a landed orchestrator mint by its `Minted` log,
+    /// full-matching `(to, nonce, token, amount)`. The on-chain uniqueness
+    /// key is only `(to, nonce)`, so a bare hit is not sufficient proof that
+    /// *this* mint landed — a hit whose token or amount differs means the
+    /// nonce was consumed by a different mint, and the caller must fail
+    /// manually rather than complete the wrong mint.
+    #[allow(dead_code)]
+    async fn find_orchestrator_minted_log(
```
0xgleb, line 330: should fix: this takes `(orchestrator, to, nonce, token, amount)` while `validate_mint_authorization` below takes `token` before `to`. Both are `Address`, so transposing them at a call site compiles cleanly and silently queries or validates the wrong pair. Either align the order across the two, or wrap them in named types so the compiler catches it.

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+        // orchestrator's `authorizeMint` callback — there is no signature to
+        // recover (SPEC Decision 1).
+        if authorization.signature.is_empty() {
+            return Ok(());
+        }
+
+        // The contract's own digest view guarantees the exact EIP-712 domain
+        // and struct hash the orchestrator will verify at mint time.
+        let digest = contract
+            .mintAuthDigest(token, to, amount, authorization.nonce)
```
0xgleb: should fix: this passes two same-typed `Address` values positionally, and nothing here pins that order to the contract's. If those two parameters are ever reordered on-chain the digest changes silently and every signature stops recovering to the expected signer, with no compile error anywhere. A test asserting the digest against a known-good fixture from the deployed contract would pin it.

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+            .call()
+            .await?;
+
+        if self.provider.get_code_at(to).await?.is_empty() {
+            let signature =
+                Signature::from_raw(authorization.signature.as_ref())?;
```
0xgleb, line 1372: should fix: this fails on a wrong length or a bad `v`, and the diff adds `VaultError::Signature` for exactly that, but nothing drives it. These bytes arrive over the authorization channel, so malformed input is the expected case rather than an exotic one. Add a case with a truncated signature asserting the typed error.

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+            let recovered = signature.recover_address_from_prehash(&digest)?;
```
0xgleb, line 1373: minor: this accepts a high-s signature that a Solidity `ecrecover` guard would reject, so a signature could pass preflight here and still fail on-chain. I have not checked what the orchestrator actually enforces, so this may well be a non-issue. Worth confirming against the contract, and normalizing s here if it does enforce low-s.

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+        // transport failure.
+        match IERC1271::new(to, &self.provider)
+            .isValidSignature(digest, authorization.signature.clone())
+            .call()
+            .await
+        {
+            Ok(magic)
+                if magic
+                    == alloy::primitives::FixedBytes(
+                        IERC1271::isValidSignatureCall::SELECTOR,
```
0xgleb, line 1398: should fix: only this wrong-magic outcome is tested. The contract-recipient branch has two further failure outcomes below, a revert reinterpreted as rejection and a genuine transport error, and neither is driven by a test. The revert-as-rejection reinterpretation is the interesting one, since it deliberately converts an `Err` into `MintAuthRejectedByContract` and nothing currently pins that.

src/vault/service.rs
```diff
@@ -863,6 +901,273 @@ impl VaultService for RealBlockchainService {
...
+                    ) =>
+            {
+                Ok(())
+            }
+            Ok(_) => Err(VaultError::MintAuthRejectedByContract { to }),
+            Err(error) if error.as_revert_data().is_some() => {
```
0xgleb, line 1403: minor: the comment says a contract that does not implement ERC-1271 is treated as a rejection, but `as_revert_data` returns `None` when a call succeeds with empty return data, which is exactly what a contract without that function does. That case falls through to the transport-error branch and propagates as an infrastructure failure rather than a rejection. Either handle empty return data explicitly or fix the comment.

src/vault/service.rs
```diff
@@ -48,6 +51,20 @@ pub type RealBlockchainServiceProvider = FillProvider<
...
 >;
 
+/// Total backward window `find_orchestrator_minted_log` scans for a landed
+/// `Minted` log, from the chain head. A consuming transaction can only land
+/// after the mint's authorization was delivered (the delivery endpoint
+/// rejects an already-consumed nonce), so the sought log is at most as old
+/// as the mint's own recovery timeline — this window (~5.8 days on Base's 2s
+/// blocks) covers the automatic retry schedule and same-week manual
+/// re-drives with ample margin.
+const MINTED_LOG_LOOKBACK_BLOCKS: u64 = 250_000;
```
0xgleb, line 62: minor: this is justified as roughly 5.8 days on Base's 2 second blocks, but the constant applies to every configured network, so a slower chain gets a much longer window and a faster one a shorter one. A landing older than the window silently returns `None` and drops to the manual path. Worth deriving it from the chain config instead of hardcoding it.

src/vault/orchestrator.rs
```diff
@@ -1,8 +1,94 @@
...
+/// `MintAuthorizationReceived`, so this serde shape is permanent.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub(crate) struct MintAuthorization {
+    /// Recipient-chosen random nonce; `(to, nonce)` is single-use on-chain.
+    pub(crate) nonce: B256,
+    /// Opaque recipient signature over the `MintAuthV1` digest. EMPTY is a
+    /// valid input: the future Atomic Bridge recipient authorizes via the
+    /// orchestrator's `authorizeMint` callback instead of a signature, so
+    /// validation skips signer recovery for empty bytes (SPEC Decision 1).
+    pub(crate) signature: Bytes,
```
0xgleb, line 24: should fix: two things about this field. An empty value is a sentinel meaning the bridge recipient authorizes via the `authorizeMint` callback, so a second authorization mode is encoded as the absence of data, and nothing in the type stops an empty signature reaching the EOA path by accident. Separately, the struct has public fields and a derived `Deserialize`, so an unvalidated authorization is representable and persistable anywhere in the crate. An enum with one variant per mode, built through a fallible constructor, would make both invalid states unrepresentable, which matters more than usual here because this shape is meant to be permanent for a future event.

src/vault/mock.rs
```diff
@@ -1513,23 +1655,236 @@ impl VaultService for MockVaultService {
...
+            // NOT a match — it is the "nonce consumed by a different mint"
+            // case.
+            return Ok(self
+                .orchestrator
+                .minted_log
+                .lock()
+                .unwrap()
+                .clone()
+                .filter(|log| {
+                    log.nonce == nonce && log.shares_minted == amount
```
0xgleb: minor: this matches on nonce and amount only, while the real implementation full-matches nonce, token, `to` and amount. Upstack tests written against this mock will green-path a different-token or different-recipient collision that production correctly rejects. Tightening it here keeps those tests honest.

src/vault/mock.rs
```diff
@@ -1969,4 +2324,250 @@ mod tests {
...
+            external_tx_id: None,
+        }
+    }
+
+    #[tokio::test]
+    async fn orchestrator_mint_mock_round_trip_records_and_confirms() {
+        let mock = MockVaultService::new_success();
+        let params = test_orchestrator_mint_params();
+
+        assert_eq!(mock.orchestrator_mint_preparation_call_count(), 0);
```
0xgleb, line 2467: minor: this asserts a vault-direct parity for `InvalidPreparedMint` that does not exist on that path, so it documents a behaviour rather than verifying one. It will keep passing regardless of whether the parity is ever introduced, which makes it worse than no test here.

src/redemption/mod.rs
```diff
@@ -391,7 +391,13 @@ const fn burn_failure_classification(
             OrchestratorRevertReason::ReceiptLogicMismatch => {
                 BurnFailureClassification::ReceiptLogicMismatch
             }
-            OrchestratorRevertReason::Unknown => {
+            // Mint-side reverts can never arise from a burn transaction;
+            // if one somehow surfaces here it stays unclassified.
+            OrchestratorRevertReason::NonceReplayed { .. }
+            | OrchestratorRevertReason::BadRecipientSignature
+            | OrchestratorRevertReason::RecipientCallbackRejected { .. }
+            | OrchestratorRevertReason::VaultAmountMismatch { .. }
```
0xgleb, line 417: minor: folding `VaultAmountMismatch` in with the other mint-side reasons rests entirely on the comment's claim that mint-side reverts can never arise from a burn. If that selector can ever surface on a burn it means the 1:1 vault ratio is broken, and `Unclassified` puts it on the auto-retry path rather than alerting. I could not settle whether the burn path can produce it, so worth confirming the selector really is mint-only.

tests/orchestrator_redemption.rs
```diff
@@ -163,26 +163,19 @@ async fn wait_for_recovery_to_resolve(
...
-/// classifications are never auto-retried. Its persistent `/admin/stuck`
-/// entry proves the seeds replay and the stuck surfacing works, so a target
-/// redemption's absence in the same snapshot is positive evidence of
-/// recovery (see `wait_for_recovery_to_resolve`).
+/// Event stream for a control redemption parked in a classified
+/// `BurnFailed` — a state recovery provably never resolves (the reconciler
+/// skips typed classifications) — seeded alongside a recovery target so the
+/// absence-based assertion in [`wait_for_recovery_to_resolve`] is
+/// non-vacuous.
 fn classified_control_events(
```
0xgleb, line 171: minor: this is a burn-path test and the changes here are a refactor unrelated to the mint support the rest of the PR adds, including weakening this helper's return from a fixed-size array to a `Vec` when no caller needs the flexibility. It makes the diff larger than the change it describes. Worth splitting out, or at least reverting the array to `Vec` part.

0xgleb (changes requested):

src/redemption/mod.rs
rouzwelt: Couldn't confirm from the ABI alone either, which is why the code already hedges ("if one somehow surfaces here it stays unclassified") rather than claiming impossibility. If it ever did surface on a burn, Unclassified → bounded retries → exhaustion → manual is the safe default for an unconfirmed case — a typed classification invented without contract confirmation would be worse. Happy to reclassify once st0x.deploy confirms the selector's reachability from burn().

src/vault/mock.rs
rouzwelt: The mock documents exactly this limitation at the site ("the fields the stored log carries — it has no to/token"). The four-field contract is pinned where it's real: the service's own find_orchestrator_minted_log_requires_full_match unit tests and the nonce-collision e2e against Anvil. Tightening the mock means adding expected-to/token state that every upstack test constructor then has to thread through, for scenarios those tests already control by nonce/amount.

src/vault/mock.rs
rouzwelt: Verified the claim first: vault-direct prepare_mint_tx only handles PrepareTxFails, so the "same way it corrupts the vault-direct one" framing was false. Doc now states what the test actually verifies: InvalidPreparedMint persists a hash that doesn't match the envelope's bytes, exercising PreparedMintTx::validate rejection. The assertion itself was always real and is unchanged.

src/vault/mod.rs
rouzwelt: Production callers land with the orchestrator mint aggregate arms stacked above this commit; the `dead_code` allows on these four methods are removed there, once the callers exist.

src/vault/orchestrator.rs
rouzwelt: The struct deliberately mirrors the on-chain MintAuthV1 tuple one-to-one — the permanent event persists exactly what is signed and submitted, and to_binding stays trivial. The mode invariant is enforced at the validation boundary (now including the empty-signature-EOA rejection from this review), which is where an unvalidated authorization is stopped from being stored. An enum would decouple the persisted shape from the contract shape it must reproduce verbatim, and ripples across five stacked branches.

src/vault/service.rs
rouzwelt: Deliberately deferred: the orchestrator rollout is Base-only, and the window's failure mode is fail-safe (an out-of-window landing drops to the documented manual path, never a false completion). Deriving from chain config is the right move when a second network gets an orchestrator.

src/vault/service.rs
rouzwelt: The scan structurally can't detect all near-misses: token and to are topic filters, so a different-token consumption never even reaches the client-side match. The authoritative "consumed" signal is nonceUsed, and the upstack callers already use it: a consumed nonce with no full match escalates to the classified NonceConsumedByOtherMint park (never auto-retried), via the confirm-side recheck. A scan-level distinct error would duplicate that contract with weaker evidence.

src/vault/service.rs
rouzwelt: The ordering is already pinned end-to-end: the e2e (orchestrator_mint_end_to_end, upstack) signs over the deployed contract's own mintAuthDigest and completes a real on-chain mint — a transposition here would fail signature verification in that test. A static fixture would pin our call against our own assumption; the e2e pins it against the contract.

src/vault/service.rs
rouzwelt: Confirmed direction matters here: normalizing s would accept signatures a low-s-enforcing contract rejects — the wrong direction. As-is, a high-s signature that the orchestrator rejects fails on-chain and classifies (fail-safe). Holding this until st0x.deploy confirms what the contract enforces; if it enforces low-s, the right change is to reject high-s at validation.

tests/orchestrator_redemption.rs
rouzwelt: Fair on diff hygiene, but this hunk is now merged history: three stacked branches build on the Vec shape (the crash-recovery test and the mint e2e both consume it via slices). Reverting to the array re-opens a conflict cascade through those branches for purely cosmetic gain, so leaving it as-is.

rouzwelt:
@coderabbit review

src/vault/service.rs
```diff
@@ -877,6 +924,323 @@ impl VaultService for RealBlockchainService {
...
+        query: MintedLogQuery,
+        authorization: &MintAuthorization,
+    ) -> Result<(), VaultError> {
+        let MintedLogQuery { orchestrator, to, nonce, token, amount, .. } =
+            query;
+        let contract = IST0xOrchestratorV1::new(orchestrator, &self.provider);
+
+        // Consumed-nonce first: definitive regardless of recipient type, and
+        // a submission with a used nonce could only revert NonceReplayed.
+        if contract.nonceUsed(to, nonce).call().await? {
```
0xgleb, line 1348: minor: Keep the query and authorization nonce tied together explicitly in `validate_mint_authorization`. Current callers construct `MintedLogQuery.nonce` from `authorization.nonce`, so this is not an exploitable preflight bypass on this head, but an assertion or single nonce source would protect the invariant from future refactors.

src/vault/mock.rs
```diff
@@ -1513,23 +1658,234 @@ impl VaultService for MockVaultService {
...
+            // Mirror the real scan's three-way verdict on the fields the
+            // stored log carries (it has no `to`/`token`): a stored log
+            // whose nonce equals the query's is the pair's one landing —
+            // matching amount is this mint's (`FullMatch`), a differing
+            // amount proves a different mint consumed the pair
+            // (`Mismatch`) — while a missing or different-nonce log means
+            // nothing was found at the pair (`NotFound`).
+            let stored = self.orchestrator.minted_log.lock().unwrap().clone();
+            return Ok(match stored {
+                Some(log) if log.nonce == nonce => {
```
0xgleb, line 1909: minor: Align the mock `find_orchestrator_minted_log` matcher with production by checking `to` and `token` as well as nonce and amount. The mismatch can only make recovery tests report a false `FullMatch`; it does not change production behavior, so this is test-double hardening rather than a merge blocker.

0xgleb (approved):

JuaniRios (approved):

PR #286, opened 2026-07-22
feat: automate orchestrator burn-failed recovery with existing tx

rouzwelt:
## Motivation

Orchestrator-mode redemptions that reached `BurnFailed` with a submitted transaction id were previously deferred entirely — the reconciler logged a warning and left them in `Failed` for manual intervention. This meant any orchestrator burn that timed out before being recorded required operator involvement even when the transaction had already landed on-chain.

## Solution

Automated orchestrator burn recovery is now wired end-to-end, mirroring the existing vault-direct path:

**Mode-scoped confirmation in** **`BurnManager`:** The early-exit guard that skipped orchestrator failures with a submitted transaction is replaced by a mode-aware branch. Vault-direct failures continue through `recover_burn_failed_with_existing_tx` (receipt lookup); orchestrator failures go through the new `recover_orchestrator_burn_failed_with_existing_tx`, which calls `confirm_orchestrator_burn`, records the result as a `RecordExistingBurn` with an `Orchestrator` proof, or terminalizes via `MarkFailed` on a definitive revert. Non-terminal errors (RPC blips, pending receipts) leave the redemption in `BurnFailed` with the tx id intact so the next pass retries confirmation rather than resubmitting.

**Mode-scoped confirmation in the admin recovery endpoint:** `inspect_prior_burn` now dispatches on `metadata.burn_mode` rather than always calling `vault_service.check_tx`. Orchestrator redemptions confirm via `confirm_orchestrator_burn`; vault-direct redemptions continue using the receipt path. Shared logic (`record_existing_burn`, `resume_after_reverted_burn`, `ambiguous_prior_burn_status`) is extracted into standalone helpers to eliminate duplication.

**`ExistingBurnProof`** **enum:** `RecordExistingBurn` now carries an `ExistingBurnProof` variant (`VaultDirect { burns }` or `Orchestrator { shares_burned, burn_range, dust_retained }`) instead of a flat `planned_burns` list. The command handler cross-checks the proof variant against the `Failed` state's anchored `burn_mode` and returns `BurnModeMismatch` if they disagree, preventing a vault-direct redemption from ever being recorded with an orchestrator proof or vice versa.

**`OrchestratorBurnRecovered`** **event:** A new terminal event carries the orchestrator-specific fields (`shares_burned`, `burn_range`, `dust_retained`) and transitions the aggregate to `Completed`, matching the audit shape of `OrchestratorTokensBurned`. `RedemptionView` and the terminal-event dispatch table are updated accordingly.

[...]

PR #285, opened 2026-07-22
Scope burn verification by `BurnProofKind` for orchestrator vs vault-direct modes

rouzwelt:
## Motivation

Burn verification was mode-agnostic: `verify_burn_in_receipt` and `VaultService::verify_burn_tx` always looked for a `Transfer(owner -> 0x0)` event, which is the vault-direct proof shape. Orchestrator burns have a structurally different two-leg shape — `Transfer(owner -> orchestrator)` (the `transferFrom` pull) followed by `Transfer(orchestrator -> 0x0)` (the burn) — so verifying them against the vault-direct shape always fails. Additionally, alternate-burn recovery in `BurnManager::force_complete_burn` was comparing orchestrator redemptions against a per-receipt plan that doesn't exist for orchestrator mode, and was unconditionally calling `settle_reserved_burn` even though orchestrator redemptions never reserve receipts.

## Solution

Introduces `BurnProofKind` (in `vault/orchestrator.rs`) with two variants — `VaultDirect` and `Orchestrator { address }` — and threads it through `verify_burn_in_receipt`, `VaultService::verify_burn_tx`, and `BurnManager::force_complete_burn`.

`verify_burn_in_receipt` now dispatches on `BurnProofKind`: vault-direct verification is unchanged; orchestrator verification accumulates the pull leg (`Transfer(owner -> orchestrator)`) and the burn leg (`Transfer(orchestrator -> 0x0)`) separately and rejects the proof unless the pull total equals the burn total and neither is zero. A vault-direct-shaped transaction is rejected under an orchestrator proof, and vice versa.

`force_complete_burn` now branches on `metadata.burn_mode` when an alternate proving hash is supplied. Vault-direct alternate proofs are still validated against the per-receipt plan. Orchestrator alternate proofs are validated by checking that the verified nonce matches the persisted transaction's nonce, the verified shares equal the redemption's `alpaca_quantity` converted to 18-decimal share-wei, and no owner share transfers are present (orchestrator mode returns no dust on-chain). `settle_reserved_burn` is skipped entirely for orchestrator redemptions.

`BurnProofKind` is always derived from the redemption's own persisted `burn_mode` via `From<VaultMode>`, never re-resolved from the asset's current vault mode, so the proof kind stays authoritative through the incremental per-asset cutover period.

The mock records the last `BurnProofKind` passed to `verify_burn_tx` for assertion in tests.  
  
[...]

PR #280, opened 2026-07-20
feat: add `IntendBurn`/`prepare_orchestrator_burn_tx` for orchestrator persist-before-broadcast

rouzwelt:
## Motivation

Orchestrator-mode redemption burns previously lacked the persist-before-broadcast discipline that vault-direct burns use. Without a dedicated `IntendBurn` step that signs and persists the exact transaction bytes before any broadcast, a crash between signing and submission could leave a redemption in an unrecoverable state. Additionally, the pre-submit readiness check (`check_orchestrator_burn_readiness`) was not wired in as a formal gate before signing, meaning deterministic failures like `InsufficientReceipts` would only surface as unclassified preparation errors at gas estimation time.

## Solution

Introduces `IntendBurn` / `BurnIntended` as the orchestrator-mode equivalent of the vault-direct persist step. `prepare_orchestrator_burn_tx` builds and signs the exact `ST0xOrchestrator.burn()` transaction and persists it in `BurnIntended` (with `planned_burns: vec![]`) before any broadcast. `BurnTokens` then broadcasts those exact persisted bytes via `submit_orchestrator_burn`, emitting `OrchestratorBurnSubmitted`. The existing `(hash, nonce)` classify/rebroadcast/replace recovery machinery applies to orchestrator burns unchanged.

`check_orchestrator_burn_readiness` is now a formal pre-submit gate that evaluates, in order: ERC-20 allowance, `vaultLogicIsExpected()` health, and an `eth_call` simulation of the burn to classify a deterministic `InsufficientReceipts` revert before anything is signed.

`IntendBurn`/`BurnTokens` carry a mode-specific `BurnParams` enum (`VaultDirect { vault, burns, dust_shares, owner }` | `Orchestrator { token, amount, owner }`), and the handler rejects params whose mode does not match the redemption's persisted `burn_mode` anchor (`BurnModeMismatch`).

`Reprocessed` and `BurnResumed` events gain the same additive `#[serde(default)]` `burn_mode` field as `RedemptionDetected`, preserving the mode anchor across a reset or resume so a reprocessed orchestrator redemption cannot silently replay as vault-direct.

The inline `matches_persisted_burn` logic in the recovery path is extracted into a standalone function, `vault_error_to_redemption` is updated to take a reference, and several minor clippy fixes are included (`const fn`, method reference, redundant clone).  
  
Closes [RAI-1218](https://linear.app/makeitrain/issue/RAI-1218/3-burn-redemptions-via-orchestratorburn)

## Checks

[...]

PR #279, opened 2026-07-20
feat: classify `InsufficientReceipts` via burn simulation before submission

rouzwelt:
## Motivation

When an orchestrator-mode burn is attempted but the orchestrator's receipt pool cannot cover the requested amount, the transaction would revert deterministically with `InsufficientReceipts`. Previously this revert was only discoverable after gas estimation rejected the transaction at signing, surfacing as an unclassified preparation failure with no structured error state and no clear recovery path.

## Solution

A burn simulation (`eth_call`) is now performed as the final gate in `check_orchestrator_burn_readiness`, after the allowance and `vaultLogicIsExpected()` checks. If the simulation reverts with `InsufficientReceipts`, the new `OrchestratorBurnReadiness::InsufficientReceipts { shortfall }` variant is returned before any signing occurs. A `VaultLogicMismatch` or `ReceiptLogicMismatch` revert from the simulation is folded into the existing `VaultLogicMismatch` arm; any other revert propagates as an error.

In `BurnManager`, the new readiness variant is handled by recording a `BurnFailureClassification::InsufficientReceipts` failure and returning a non-retryable `BurnManagerError`. The failure is never auto-retried — recovery requires a manual `EMERGENCY_ROLE` action followed by an admin `ResumeBurn`.

Four end-to-end tests on Anvil cover the full orchestrator redemption surface:

- **`orchestrator_burn_walks_multiple_receipts`** — a burn spanning three orchestrator-custodied receipts completes in a single `burn()` call and advances the per-token receipt pointer correctly.
- **`orchestrator_shortfall_is_classified_without_submission`** — a redemption larger than the receipt pool is classified as `InsufficientReceipts`, parked in `BurnFailed`, and never submitted even after multiple reconciler passes.
- **`orchestrator_recovery_confirms_landed_burn_without_resubmitting`** — startup recovery of a burn that already landed on-chain confirms the existing transaction rather than submitting a second one.
- **`mixed_mode_assets_each_take_their_own_burn_path`** — an orchestrator-mode asset and a vault-direct asset redeem concurrently, each following its own burn path without interference.

Two harness helpers (`orchestrator_mint_to`, `approve_orchestrator`) and a `create_config_with_vault_modes` convenience constructor are added to support these tests.  
  
Closes [RAI-1489](https://linear.app/makeitrain/issue/RAI-1489/e2e-anvil-tests)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
[...]

PR #298, opened 2026-07-27
feat: orchestrator mint authorization endpoint (RAI-1243)

rouzwelt:
## Motivation

Orchestrator-mode mints require a `MintAuthV1` EIP-712 signature from the liquidity bot before the on-chain transaction can be submitted. Alpaca cannot carry this authorization through its flow, so it must arrive out-of-band on an internal channel. Without a dedicated endpoint and aggregate command to receive and persist it, orchestrator-mode mints have no way to acquire the authorization they need before the `PrepareMint` step.

Additionally, the `VaultMode` for a mint was previously derived from live config at each step, meaning a config change mid-flight could alter the behavior of an in-progress mint. The mode needs to be anchored at initiation time and carried through the entire lifecycle.

## Solution

A `mint_mode` field (`VaultDirect` | `Orchestrator { address }`) is resolved from config exactly once at `Initiate` time and persisted on the `Initiated` event and all subsequent lifecycle states. Every mode-dependent step derives from this persisted anchor rather than live config. Historical events and snapshots that predate this field default to `VaultDirect` via `#[serde(default)]`.

A `mint_authorization` field (`Option<MintAuthorization>`) is added to all pre-terminal `Mint` states to carry the liquidity bot's validated `MintAuthV1` nonce and signature. It is always `None` on a freshly initiated mint and populated by the new `MintAuthorizationReceived` event.

A new `POST /internal/mints/{tokenization_request_id}/authorization` endpoint (`authorize_mint`) accepts the liquidity bot's authorization delivery. It:

- Looks up the mint by `tokenization_request_id` (the only mint identifier shared with the liquidity bot)
- Rejects delivery for vault-direct mints with `422`
- Validates the authorization on-chain (signer identity, nonce consumption) before touching the aggregate, so a bad delivery is an actionable failure at this call rather than a post-journal surprise
- Issues `MintCommand::AuthorizeMint`, which produces `MintAuthorizationReceived` without changing the lifecycle state
- Is idempotent on redelivery of an identical authorization and returns `409` for a conflicting nonce

The aggregate enforces that authorization is only accepted in `Initiated`, `JournalConfirmed`, and `Minting` states — once `PrepareMint` signs, the nonce is baked into the persisted transaction bytes and a late delivery cannot change what gets submitted.  
  
Closes [RAI-1616](https://linear.app/makeitrain/issue/RAI-1616/orchestrator-mint-authorization-endpoint)

## Checks

## Summary by CodeRabbit

* **New Features**
  * Added mint authorization with validation, recording, and idempotent redelivery handling.
  * Mint workflows now preserve vault mode and authorization details throughout their lifecycle.
  * Added tokenization-request lookup to route authorizations to the correct mint.
* **Bug Fixes**
  * Improved handling of stale, duplicate, closed, unsupported, and ambiguous authorization requests.
  * Prevented orchestrator-anchored mints from using the direct-vault submission flow.
* **Documentation**
  * Documented the mint authorization endpoint and its request and response formats.

### DIFF
--- a/src/mint/api/authorize.rs
+++ b/src/mint/api/authorize.rs
@@ -1,0 +1,1001 @@
+use alloy::primitives::{B256, Bytes};
+use cqrs_es::AggregateError;
+use event_sorcery::{LifecycleError, Store};
+use rocket::http::{ContentType, Status};
+use rocket::post;
+use rocket::response::{self, Responder};
+use rocket::serde::json::Json;
+use serde::{Deserialize, Serialize};
+use sqlx::{Pool, Sqlite};
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::timeout;
+use tracing::{error, info, warn};
+
+use super::ErrorResponse;
+use crate::auth::InternalAuth;
+use crate::config::VaultMode;
+use crate::mint::view::find_issuer_id_by_tokenization_request_id;
+use crate::mint::{
+    IssuerMintRequestId, Mint, MintCommand, MintError, TokenizationRequestId,
+};
+use crate::tokenized_asset::view::find_vault;
+use crate::vault::{
+    MintAuthorization, MintedLogQuery, NetworkVaultServices, VaultError,
+};
+
+/// Wire shape of the liquidity bot's mint-authorization delivery
+/// (RAI-1243). The signature is opaque bytes; `"0x"` (empty) is valid input
+/// for recipients authorized via the orchestrator's `authorizeMint`
+/// callback.
+#[derive(Debug, Deserialize, utoipa::ToSchema)]
+pub(crate) struct MintAuthorizationRequest {
+    /// Recipient-chosen random 32-byte nonce, hex-encoded.
+    #[schema(value_type = String)]
+    pub(crate) nonce: B256,
+    /// EIP-712 `MintAuthV1` signature over `(token, to, amount, nonce)` by
+    /// the recipient wallet key, hex-encoded; may be `"0x"`.
+    #[schema(value_type = String)]
+    pub(crate) signature: Bytes,
+}
+
+#[derive(Debug, Serialize, utoipa::ToSchema)]
+pub(crate) struct MintAuthorizationResponse {
+    #[schema(value_type = String)]
+    pub(crate) issuer_request_id: IssuerMintRequestId,
+    pub(crate) status: String,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum MintAuthorizationApiError {
+    #[error("No mint found for tokenization request {0}")]
+    UnknownTokenizationRequest(TokenizationRequestId),
+
+    #[error("Mint {0} is vault-direct; it never consumes an authorization")]
+    VaultDirectMint(TokenizationRequestId),
+
+    #[error("Authorization not acceptable: mint is in state {0}")]
+    NotAcceptable(String),
+
+    #[error("A different authorization is already recorded for this mint")]
+    ConflictingAuthorization,
+
+    #[error("The mint was modified concurrently; retry the delivery")]
+    DeliveryRace,
+
+    #[error("Invalid authorization: {0}")]
+    InvalidAuthorization(VaultError),
+
+    #[error("On-chain validation failed: {0}")]
+    OnChainValidationFailed(VaultError),
+
+    #[error("On-chain validation timed out")]
+    OnChainValidationTimedOut,
+
+    #[error("Internal error")]
+    Internal,
+}
+
+/// Upper bound on the on-chain reads validating one authorization (up to
+/// four sequential RPCs: `nonceUsed`, `mintAuthDigest`, `eth_getCode`, and
+/// ERC-1271 `isValidSignature`). The HTTP provider has no request timeout of
+/// its own, so without this deadline an unresponsive RPC node would hold the
+/// liquidity bot's delivery request open indefinitely.
+#[cfg(not(test))]
+const ON_CHAIN_VALIDATION_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Unit tests hang the mock validation forever to exercise the deadline; a
+/// millisecond-scale bound keeps that test fast in real time. A paused
+/// tokio clock is not an option here — auto-advance also fires sqlx's
+/// pool-acquire timers, failing the request with `PoolTimedOut` before
+/// validation is ever reached whenever an acquire has to wait.
+#[cfg(test)]
+const ON_CHAIN_VALIDATION_TIMEOUT: Duration = Duration::from_millis(50);
+
+impl<'r> Responder<'r, 'static> for MintAuthorizationApiError {
+    fn respond_to(
+        self,
+        _req: &'r rocket::Request<'_>,
+    ) -> response::Result<'static> {
+        let (status, message) = match &self {
+            Self::UnknownTokenizationRequest(_) => {
+                (Status::NotFound, self.to_string())
+            }
+            Self::VaultDirectMint(_) | Self::InvalidAuthorization(_) => {
+                (Status::UnprocessableEntity, self.to_string())
+            }
+            Self::NotAcceptable(_)
+            | Self::ConflictingAuthorization
+            | Self::DeliveryRace => (Status::Conflict, self.to_string()),
+            // Unlike the deliberately descriptive 422s, the read-failure
+            // body stays generic: the underlying `VaultError` can carry
+            // transport/provider detail (RPC endpoints, connection errors)
+            // that must not leave the process. The full error is already
+            // ERROR-logged where this variant is constructed.
+            Self::OnChainValidationFailed(_)
+            | Self::OnChainValidationTimedOut => (
+                Status::BadGateway,
+                "On-chain validation is currently unavailable".to_string(),
+            ),
+            Self::Internal => (
+                Status::InternalServerError,
+                "Internal server error".to_string(),
+            ),
+        };
+
+        let body = serde_json::to_string(&ErrorResponse { error: message })
+            .map_err(|_| Status::InternalServerError)?;
+        rocket::Response::build()
+            .status(status)
+            .header(ContentType::JSON)
+            .sized_body(body.len(), std::io::Cursor::new(body))
+            .ok()
+    }
+}
+
+/// Receives the liquidity bot's `MintAuthV1` for one orchestrator-mode mint,
+/// validates it on-chain against the mint's own persisted facts
+/// (`token`, `to`, `amount` — the bot MUST mint with exactly the signed
+/// values), and records it on the aggregate. Delivered out-of-band from the
+/// Alpaca flow: Alpaca cannot carry the authorization, so it arrives on the
+/// same internal channel the liquidity bot already uses for the asset-status
+/// endpoint.
+#[utoipa::path(
+    post,
+    path = "/internal/mints/{tokenization_request_id}/authorization",
+    tag = "internal",
+    params(
+        ("tokenization_request_id" = String, Path,
+            description = "Alpaca tokenization request id the mint was initiated with")
+    ),
+    request_body = MintAuthorizationRequest,
+    responses(
+        (status = 200, description = "Authorization validated and recorded (idempotent)",
+            body = MintAuthorizationResponse),
+        (status = 404, description = "No mint for this tokenization request"),
+        (status = 409,
+            description = "Conflicting authorization, or the mint already signed its transaction"),
+        (status = 422,
+            description = "Vault-direct mint, invalid or malformed signer, \
+                empty signature for an EOA recipient, or consumed nonce"),
+        (status = 502, description = "On-chain validation read failure")
+    ),
+    security(("internal_api_key" = []))
+)]
+#[tracing::instrument(skip(_auth, mint_store, pool, vault_services, request))]
+#[post(
+    "/internal/mints/<tokenization_request_id>/authorization",
+    format = "json",
+    data = "<request>"
+)]
+pub(crate) async fn authorize_mint(
+    _auth: InternalAuth,
+    tokenization_request_id: &str,
+    mint_store: &rocket::State<Arc<Store<Mint>>>,
+    pool: &rocket::State<Pool<Sqlite>>,
+    vault_services: &rocket::State<NetworkVaultServices>,
+    request: Json<MintAuthorizationRequest>,
+) -> Result<Json<MintAuthorizationResponse>, MintAuthorizationApiError> {
+    let tokenization_request_id =
+        TokenizationRequestId(tokenization_request_id.to_string());
+    let request = request.into_inner();
+
+    let issuer_request_id = find_issuer_id_by_tokenization_request_id(
+        pool.inner(),
+        &tokenization_request_id,
+    )
+    .await
+    .map_err(|err| {
+        error!(target: "mint", error = %err,
+            "Failed to look up mint by tokenization request id"
+        );
+        MintAuthorizationApiError::Internal
+    })?
+    .ok_or_else(|| {
+        warn!(target: "mint",
+            tokenization_request_id = %tokenization_request_id,
+            "Authorization delivered for an unknown tokenization request"
+        );
+        MintAuthorizationApiError::UnknownTokenizationRequest(
+            tokenization_request_id.clone(),
+        )
+    })?;
+
+    let mint = mint_store
+        .load(&issuer_request_id)
+        .await
+        .map_err(|err| {
+            error!(target: "mint", issuer_request_id = %issuer_request_id,
+                error = %err, "Failed to load mint aggregate"
+            );
+            MintAuthorizationApiError::Internal
+        })?
+        .ok_or_else(|| {
+            error!(target: "mint", issuer_request_id = %issuer_request_id,
+                "Mint aggregate missing for known tokenization request"
+            );
+            MintAuthorizationApiError::Internal
+        })?;
+
+    // A mint with no live mode (`Closed`) is not a vault-direct mint — it
+    // cannot accept an authorization in any mode, so it must not take the
+    // vault-direct rejection below with its untrue cause.
+    let Some(mode) = mint.mint_mode() else {
+        return Err(MintAuthorizationApiError::NotAcceptable(
+            mint.state_name().to_string(),
+        ));
+    };
+
+    // The orchestrator address comes from the mint's own persisted
+    // `mint_mode` anchor — never live config. A vault-direct mint has no
+    // orchestrator to validate against and never consumes an authorization.
+    let VaultMode::Orchestrator { address: orchestrator } = mode else {
+        warn!(target: "mint", issuer_request_id = %issuer_request_id,
+            "Authorization delivered for a vault-direct mint; rejecting"
+        );
+        return Err(MintAuthorizationApiError::VaultDirectMint(
+            tokenization_request_id.clone(),
+        ));
+    };
+
+    let (Some(network), Some(underlying), Some(to), Some(quantity)) =
+        (mint.network(), mint.underlying(), mint.wallet(), mint.quantity())
+    else {
+        return Err(MintAuthorizationApiError::NotAcceptable(
+            mint.state_name().to_string(),
+        ));
+    };
+
+    let amount = quantity.to_u256_with_18_decimals().map_err(|err| {
+        error!(target: "mint", issuer_request_id = %issuer_request_id,
+            error = %err, "Persisted mint quantity failed share conversion"
+        );
+        MintAuthorizationApiError::Internal
+    })?;
+
+    let vault = find_vault(pool.inner(), underlying, &network)
+        .await
+        .map_err(|err| {
+            error!(target: "mint", issuer_request_id = %issuer_request_id,
+                error = %err, "Vault lookup failed"
+            );
+            MintAuthorizationApiError::Internal
+        })?
+        .ok_or_else(|| {
+            error!(target: "mint", issuer_request_id = %issuer_request_id,
+                underlying = %underlying.as_str(),
+                "No vault for the mint's asset"
+            );
+            MintAuthorizationApiError::Internal
+        })?;
+
+    let vault_service = vault_services.service(network).map_err(|err| {
+        error!(target: "mint", issuer_request_id = %issuer_request_id,
+            error = %err,
+            "No vault service for the mint's network"
+        );
+        MintAuthorizationApiError::Internal
+    })?;
+
+    let authorization = MintAuthorization {
+        nonce: request.nonce,
+        signature: request.signature,
+    };
+
+    // An identical redelivery is the common retry case (a delivery whose
+    // response was lost): answer from the recorded state without re-running
+    // the on-chain validation — the recorded authorization already passed
+    // it, and re-validating burns up to four RPC reads per retry.
+    if mint.accepts_mint_authorization()
+        && mint.mint_authorization() == Some(&authorization)
+    {
+        info!(target: "mint", issuer_request_id = %issuer_request_id,
+            tokenization_request_id = %tokenization_request_id,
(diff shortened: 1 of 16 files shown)


rouzwelt:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/298?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#298** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/298?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/298?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=298).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

rouzwelt:
@coderabbit review

src/mint/mod.rs
```diff
@@ -554,6 +674,174 @@ impl Mint {
...
+        match existing {
+            Some(existing) if *existing == mint_authorization => Ok(vec![]),
+            Some(_) => Err(MintError::ConflictingMintAuthorization),
+            None => Ok(vec![MintEvent::MintAuthorizationReceived {
+                issuer_request_id: provided_id,
+                mint_authorization,
+                received_at: Utc::now(),
+            }]),
+        }
+    }
```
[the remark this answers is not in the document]
rouzwelt: The submit job never signs an orchestrator mint without its authorization — it defers with 'awaiting authorization' until one is recorded (job.rs orchestrator arm), so no signed bytes can omit the nonce. Minting must keep accepting delivery because Deposit auto-advances on journal confirmation before the bot delivers; rejecting there would strand normal mints. Nonce swap mid-flight is blocked by ConflictingMintAuthorization + idempotent redelivery, and post-intent states reject/no-op."

src/mint/api/initiate.rs
```diff
@@ -64,6 +66,12 @@ pub(crate) async fn initiate_mint(
 
     let issuer_request_id = IssuerMintRequestId::random();
 
+    // Resolve the asset's mode from config exactly once, here at initiate
+    // time: the persisted anchor on `Initiated` is what every later
+    // mode-dependent step derives from, even if the asset's configured
+    // vault_mode flips mid-flight.
+    let mint_mode = config.vault_mode_for(&request.underlying);
```
0xgleb, line 73: should fix: as of this diff the persisted `mint_mode` anchor is written but nothing on the mint execution path consumes it. The `Minting` handler in mint/job.rs calls `prepare_mint_tx` and `submit_mint`, the vault-direct multicall, with no branch on `mint_mode`, and `handle_record_tx_intended` accepts a vault-direct `PreparedMintTx` for an Orchestrator-anchored mint. The redemption side fails closed on the mirror case with `BurnModeMismatch`; here it fails open.

Stack note: #300 does close this, adding `MintError::MintModeMismatch` and rejecting orchestrator-anchored mints in `RecordTokensMinted` plus vault-direct ones in the two orchestrator commands, with tests. So this is a gap in the intermediate state rather than the end state. Worth confirming the two land together, because between them there is a commit where flipping an asset to orchestrator mode answers the bot 200 authorized and then mints vault-direct anyway.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+        VaultError::MintAuthSignerMismatch { .. }
+        | VaultError::MintAuthNonceUsed { .. }
+        | VaultError::MintAuthRejectedByContract { .. }
+        | VaultError::Signature(_) => {
+            warn!(target: "mint", issuer_request_id = %issuer_request_id,
+                error = %err, "Rejected invalid mint authorization"
+            );
+            MintAuthorizationApiError::InvalidAuthorization(err)
+        }
+        _ => {
```
0xgleb, line 339: should fix: this classifier names four `VaultError` variants as 422 and funnels the rest through a wildcard into 502, and those two buckets mean opposite things to the bot, permanently bad versus retry the same thing. That is not just a future-variant hazard, it is already wrong: `validate_mint_authorization` can return `VaultError::SignerRecovery` when the bytes parse as a signature but fail ECDSA recovery, which is permanent, and it lands in the retryable 502 bucket today. Make the match exhaustive, or put a `mint_auth_verdict` method next to the enum so a new variant fails to compile until someone picks its classification.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+    let authorization = MintAuthorization {
+        nonce: request.nonce,
+        signature: request.signature,
+    };
+
+    // Validate on receipt so a bad delivery is an actionable failure on this
+    // internal call, not a post-journal surprise at the on-chain step.
+    timeout(
+        ON_CHAIN_VALIDATION_TIMEOUT,
+        vault_service.validate_mint_authorization(
```
0xgleb, line 307: should fix: the nonce is chosen entirely by the caller and validation only asks the chain `nonceUsed(to, nonce)`, which is an already-consumed check, not a pending-elsewhere check. Nothing looks across mints for the same nonce, so two in-flight mints to the same recipient can both be authorized on one nonce and whichever lands second reverts `NonceReplayed` after the tx was already signed and broadcast. A uniqueness check against outstanding authorizations at delivery time turns a broadcast-then-revert into a clean 422.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+        _req: &'r rocket::Request<'_>,
+    ) -> response::Result<'static> {
+        let (status, message) = match &self {
+            Self::UnknownTokenizationRequest(_) => {
+                (Status::NotFound, self.to_string())
+            }
+            Self::VaultDirectMint(_) | Self::InvalidAuthorization(_) => {
+                (Status::UnprocessableEntity, self.to_string())
+            }
+            Self::NotAcceptable(_) | Self::ConflictingAuthorization => {
```
0xgleb: should fix: an optimistic-lock conflict on the mint aggregate comes back as a 500 here, while every sibling endpoint maps that same condition to 409. The bot cannot distinguish a genuine server fault from lost a race it should simply retry, so a concurrent delivery turns into an alert instead of a retry. Map the aggregate conflict to 409 like the siblings do.

src/mint/view.rs
```diff
@@ -215,6 +227,94 @@ pub(crate) async fn find_by_issuer_request_id(
...
+        if mint.accepts_mint_authorization() {
+            accepting.push(issuer_request_id);
+        } else {
+            stale.push(issuer_request_id);
+        }
+    }
+
+    match (accepting.len(), stale.len()) {
+        (1, _) => Ok(accepting.pop()),
+        (0, 0 | 1) => Ok(stale.pop()),
```
0xgleb, line 291: should fix: two problems in this match. A duplicate Alpaca issuance POST leaves two accepting rows for one tokenization request, which falls to the ambiguity arm and returns a 500 on every subsequent authorization delivery, so the mint is permanently wedged by something entirely outside our control. Separately, in the `(0, matches) | (matches, _)` arm the binding takes the first tuple element on the second alternative, so when both accepting and stale rows exist the reported count is `accepting.len()` and silently undercounts. Handle the duplicate case explicitly rather than letting it reach the ambiguity arm, and bind the real total.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+    UnknownTokenizationRequest(TokenizationRequestId),
+
+    #[error("Mint {0} is vault-direct; it never consumes an authorization")]
+    VaultDirectMint(TokenizationRequestId),
+
+    #[error("Authorization not acceptable: mint is in state {0}")]
+    NotAcceptable(String),
+
+    #[error("A different authorization is already recorded for this mint")]
+    ConflictingAuthorization,
```
0xgleb, line 61: should fix: redelivering an authorization that was already recorded is rejected rather than treated as a no-op. Delivery is a network call from another service, so a retry after a timeout that actually succeeded is the normal case, not an exotic one, and the bot gets an error for a state that is already exactly what it wanted. Make an identical redelivery idempotent and reserve the rejection for a genuinely different authorization on a recorded mint.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+            error!(target: "mint", issuer_request_id = %issuer_request_id,
+                "Mint aggregate missing for known tokenization request"
+            );
+            MintAuthorizationApiError::Internal
+        })?;
+
+    // A mint with no live mode (`Closed`) is not a vault-direct mint — it
+    // cannot accept an authorization in any mode, so it must not take the
+    // vault-direct rejection below with its untrue cause.
+    let Some(mode) = mint.mint_mode() else {
```
0xgleb, line 223: should fix: this is the guard that stops a Closed mint from being authorized, and `mint_mode()` returns `None` only for `Closed`, so the branch exists purely to reject that case. No test drives it. Given it is the boundary between a live mint and a dead one on a money path, it should have a test that closes a mint and asserts the rejection.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+        assert!(logs_contain_at!(
+            tracing::Level::WARN,
+            &["Conflicting mint authorization rejected"]
+        ));
+    }
+
+    /// A non-authorization on-chain read failure surfaces as 502 with a
+    /// generic body — never a 422 rejection, and never the underlying
+    /// transport/provider detail.
+    #[tokio::test]
```
0xgleb, line 838: should fix: several of the new tests carry `#[traced_test]` and drive a path that emits a WARN, but none of them assert on the captured output, so the attribute is doing nothing. That covers `invalid_authorization_is_unprocessable`, `closed_mint_delivery_is_not_found_never_vault_direct` and `post_intent_delivery_is_a_conflict`. These are the operator-visible signals for a rejected authorization, so either assert the log the test already set itself up to capture, or drop the attribute.

src/mint/mod.rs
```diff
@@ -146,6 +149,19 @@ pub(crate) enum Mint {
...
+        /// `VaultDirect`.
+        #[serde(default)]
+        mint_mode: VaultMode,
+        /// The liquidity bot's validated `MintAuthV1`, absent until the
+        /// internal mint-authorization call arrives (orchestrator mode only;
+        /// always `None` for vault-direct mints). Orthogonal to `mint_mode` —
+        /// an orchestrator mint whose authorization has not arrived yet is
+        /// still an orchestrator mint.
+        #[serde(default)]
+        mint_authorization: Option<MintAuthorization>,
```
0xgleb, line 166: should fix: carrying `mint_mode` and `mint_authorization` as independent fields on all nine live variants makes `VaultDirect` plus `Some(authorization)` representable, which is a state that should not exist. The invariant is currently held only by the check in `handle_authorize_mint`, so anything that constructs or deserializes a state directly can bypass it. Hanging the authorization off the Orchestrator variant of the mode would make the bad pairing unrepresentable instead of merely unreachable.

src/mint/mod.rs
```diff
@@ -554,6 +674,174 @@ impl Mint {
...
+
+        Self::validate_issuer_request_id(expected_id, &provided_id)?;
+
+        if matches!(mint_mode, VaultMode::VaultDirect) {
+            return Err(MintError::AuthorizationForVaultDirectAsset {
+                underlying: underlying.clone(),
+            });
+        }
+
+        match existing {
```
0xgleb, line 866: minor: once a wrong authorization is recorded first, the mint has no way back. The state accepts one authorization and then rejects any different one, so a bot that delivers a bad nonce and then corrects it is stuck, and the only exit is an admin path that does not exist yet. Worth either allowing a replacement while the mint is still pre-intent, or adding an explicit admin clear.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+
+    #[error("On-chain validation timed out")]
+    OnChainValidationTimedOut,
+
+    #[error("Internal error")]
+    Internal,
+}
+
+/// Upper bound on the on-chain reads validating one authorization (up to
+/// four sequential RPCs: `nonceUsed`, `mintAuthDigest`, `eth_getCode`, and
```
0xgleb, line 80: minor: every delivery runs the full on-chain validation, up to four RPC calls, before looking at whether this mint already holds an identical authorization, and the endpoint has no rate limit. A bot stuck in a retry loop turns into sustained RPC load on a provider we pay per call. Checking the recorded authorization first would short-circuit the common retry.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+
+    let (Some(network), Some(underlying), Some(to), Some(quantity)) =
+        (mint.network(), mint.underlying(), mint.wallet(), mint.quantity())
+    else {
+        return Err(MintAuthorizationApiError::NotAcceptable(
+            mint.state_name().to_string(),
+        ));
+    };
+
+    let amount = quantity.to_u256_with_18_decimals().map_err(|err| {
```
0xgleb, line 249: minor: the signed amount is derived from the quantity Alpaca sent at initiation, stored with no precision normalization, then converted with `to_u256_with_18_decimals`. The digest the recipient signs is only correct if that quantity round-trips exactly at 18 decimals. Worth confirming what precision Alpaca actually guarantees on `qty`, because a rounding difference here produces a signature that recovers to the wrong signer rather than an obvious error.

src/mint/event.rs
```diff
@@ -99,6 +109,17 @@ pub(crate) enum MintEvent {
...
     },
+
+    /// The liquidity bot's validated `MintAuthV1` for this mint, delivered
+    /// out-of-band via the internal mint-authorization call (orchestrator
+    /// mode only). This is the persistence point for the nonce — `Initiated`
+    /// is written on the Alpaca POST, strictly before the authorization
+    /// exists, so it cannot carry one. Does not change the lifecycle state.
+    MintAuthorizationReceived {
+        issuer_request_id: IssuerMintRequestId,
+        mint_authorization: MintAuthorization,
```
0xgleb, line 130: minor: the event records the nonce and signature but not the `(token, to, amount)` those were validated against, so a later reader cannot reconstruct what was actually checked without re-deriving it from the mint state and current config. If the vault address or wallet mapping ever changes, that re-derivation stops matching what was validated. Is the intent that validation is never re-checked from the event, or would carrying the tuple be cheap enough to just do?

Also on this event: the `Initiated` wire shape gains fields and the mint state schema changes, but neither `event_version()` nor `SCHEMA_VERSION` moves. If those are the replay-compatibility markers they should move with the shape; if newly added serde-default fields count as compatible by convention that is fine, but worth being deliberate given these events are replayed.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+    mint_store: &rocket::State<Arc<Store<Mint>>>,
+    pool: &rocket::State<Pool<Sqlite>>,
+    vault_services: &rocket::State<NetworkVaultServices>,
+    request: Json<MintAuthorizationRequest>,
+) -> Result<Json<MintAuthorizationResponse>, MintAuthorizationApiError> {
+    let tokenization_request_id =
+        TokenizationRequestId(tokenization_request_id.to_string());
+    let request = request.into_inner();
+
+    let issuer_request_id = find_issuer_id_by_tokenization_request_id(
```
0xgleb, line 183: minor: the endpoint is keyed on Alpaca's `tokenization_request_id`, which assumes the liquidity bot can obtain that id. Worth confirming that is actually true on the bot side, because if it only holds our issuer request id then this correlation key does not work end to end and it would be better to learn that before the endpoint ships than after.

src/mint/mod.rs
```diff
@@ -554,6 +674,174 @@ impl Mint {
...
+            | Self::Completed { quantity, .. } => Some(quantity),
+            Self::Closed { .. } => None,
+        }
+    }
+
+    /// Whether this mint's lifecycle state can still accept a recipient
+    /// authorization — the exact states [`Self::handle_authorize_mint`]
+    /// destructures (keep the two in sync). The tokenization-id lookup uses
+    /// this to prefer a live mint over stale same-id duplicates.
+    pub(crate) const fn accepts_mint_authorization(&self) -> bool {
```
0xgleb, line 805: minor: the set of authorization-accepting states is written out three separate times, in `accepts_mint_authorization`, in the `handle_authorize_mint` let-else, and in the apply function. Adding a fourth accepting state means finding all three, and missing one fails silently rather than loudly. One predicate used by the other two would collapse it.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+        (status = 409,
+            description = "Conflicting authorization, or the mint already signed its transaction"),
+        (status = 422,
+            description = "Vault-direct mint, invalid signer, or consumed nonce"),
+        (status = 502, description = "On-chain validation read failure")
+    ),
+    security(("internal_api_key" = []))
+)]
+#[tracing::instrument(skip(_auth, mint_store, pool, vault_services, request))]
+#[post(
```
0xgleb, line 166: minor: SPEC.md is not updated for the new internal endpoint or for the fact that authorization validation now lives at the API layer rather than where the spec describes it. Given the spec is the artifact the other repo integrates against, this is the change most worth reflecting in it.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+    .await
+    .map_err(|err| {
+        error!(target: "mint", error = %err,
+            "Failed to look up mint by tokenization request id"
+        );
+        MintAuthorizationApiError::Internal
+    })?
+    .ok_or_else(|| {
+        warn!(target: "mint",
+            tokenization_request_id = %tokenization_request_id.0,
```
0xgleb: nit: this logs `tokenization_request_id.0`, reaching past the newtype even though it implements `Display`. Same pattern again a little further down. Using the `Display` impl keeps the field format in one place.

src/mint/api/authorize.rs
```diff
@@ -0,0 +1,952 @@
...
+            .manage(vault_services)
+            .mount("/", routes![authorize_mint])
+    }
+
+    async fn deliver<'client>(
+        client: &'client rocket::local::asynchronous::Client,
+        tokenization_request_id: &str,
+        nonce: B256,
+        signature: &str,
+        with_key: bool,
```
0xgleb: nit: the `deliver` test helper takes a trailing `with_key: bool`, so call sites read as a bare `true` or `false` with no clue what is being toggled. The repo guidance calls this out specifically. Two named helpers, or a small enum, would make the call sites say what they mean.

0xgleb (changes requested):

src/mint/api/initiate.rs
rouzwelt: dded a fail-closed guard in job.rs: the Mint::Minting arm now destructures mint_mode and, for VaultMode::Orchestrator, WARNs and defers (mint stays Minting, visible in /admin/stuck), mirroring redemption's BurnModeMismatch refusal. The comment itself notes #300 supersedes this upstack — the guard closes the window in between. New test submit_mint_job_defers_orchestrator_anchored_mint asserts state unchanged, zero vault calls, and the WARN log.

src/mint/event.rs
rouzwelt: The tuple is intentionally not carried: it's fully derivable from the mint's own event-sourced state (quantity/underlying/wallet/network recorded at Initiated), not from current config — a later config change can't alter the reconstruction, because reconstruction never consults config. On versions: Initiated.mint_mode carries #[serde(default)] (verified), and additive defaulted fields are this repo's established replay-compatibility convention — old payloads deserialize to VaultDirect, which is correct for all pre-existing events.

src/mint/view.rs
rouzwelt: The binding bug was fixed, the dup-POST part though:
two accepting mints for one tokenization request means we cannot know which mint the authorization belongs to — silently picking one risks attaching a nonce to the wrong mint on a money path. Failing loudly (500 + AmbiguousTokenizationRequest in logs) is the deliberate fail-closed choice; the wedge is resolved by closing the duplicate mint, which is the actual defect.

src/mint/mod.rs
rouzwelt: VaultMode is a shared config type whose externally-tagged JSON is a permanent event wire format used by both mint and redemption anchoring — embedding a mint-specific authorization in it would change that permanent shape and couple the config type to one aggregate. The state is event-derived: MintAuthorizationReceived is only emitted after the mode check, so the "bad pairing" cannot be constructed through replay, and aggregates are only ever built through replay.

src/mint/mod.rs
rouzwelt: Nonce stability after recording is the invariant recovery depends on: once recorded, the nonce may already be baked into signed bytes or in flight, so allowing pre-intent replacement reintroduces exactly the mid-flight nonce swap this guard exists to prevent. The exit for a genuinely wrong authorization is CloseMint + re-initiate, which exists today.

src/mint/mod.rs
rouzwelt: The three sites can't collapse to one predicate: accepts_mint_authorization is a boolean over &self, the let-else in handle_authorize_mint destructures shared borrows of five fields, and the apply site needs &mut bindings — three different borrow shapes. The doc comments on handle_authorize_mint and apply_mint_authorization_received explicitly cross-link the lists so a divergence is caught in review; and a missed site fails safe (rejects/no-ops rather than accepting from a wrong state).

src/mint/api/authorize.rs
rouzwelt: Identical redelivery is already a no-op: handle_authorize_mint returns Ok(vec![]) when the recorded authorization matches, and only a different authorization gets ConflictingMintAuthorization — exactly the split this comment asks for. Covered by records_valid_authorization_and_redelivery_is_idempotent. (Redelivery now also short-circuits before on-chain validation — see the rate-limit comment.)

src/mint/api/authorize.rs
rouzwelt: Settled in the SPEC update upstack (#303): the correlation contract is documented there — Alpaca's tokenization request id is the id the bot holds from its side of the flow, which is why it's the endpoint key.

src/mint/api/authorize.rs
rouzwelt: The candidate lookup (find_issuer_id_by_tokenization_request_id) only resolves accepting/pre-intent mints, so a Closed mint never reaches this guard through the endpoint — it's defense-in-depth against a load-after-close race, which is not deterministically drivable from a test. The endpoint-level Closed behavior (404, never vault-direct) is covered by closed_mint_delivery_is_not_found_never_vault_direct.

src/mint/api/authorize.rs
rouzwelt: to_u256_with_18_decimals fails fast on any precision loss — it errors rather than rounds, so a quantity that doesn't round-trip exactly at 18 decimals produces an explicit conversion error at authorization time, never a digest that silently recovers to the wrong signer. Alpaca's qty precision doesn't change that failure mode.

src/mint/api/authorize.rs
rouzwelt: The nonce is generated randomly per-authorization by the bot per SPEC — a collision across two in-flight mints to the same recipient is a bot bug, not an expected flow. If it happens, the second tx reverts NonceReplayed, the failure records as Unclassified, and retry converges (the redelivered authorization re-validates against nonceUsed and gets a clean 422 then). An outstanding-authorization index would add a cross-aggregate read model to turn one already-recoverable bot bug into a slightly earlier error; not worth the surface.

0xgleb (approved):

JuaniRios (approved):

PR #277, opened 2026-07-20
feat: skip classified/orchestrator burn failures in reconciler, add mode-correct recovery params

rouzwelt:
## Motivation

The burn recovery reconciler previously assumed all unresolved burns should be retried automatically and that all recovery paths used the vault-direct `BurnParams` shape. This caused two problems:

1. Classified failures (`AllowanceInsufficient`, `InsufficientReceipts`, `VaultLogicMismatch`) are deterministic until an operator acts — auto-retrying them wastes recovery budget and can mask the root cause.
2. Orchestrator-mode burns with a submitted transaction require the orchestrator-shaped recovery path (`OrchestratorBurnRecovered`), not the vault-direct confirm path, so running the wrong path against an orchestrator transaction was incorrect.

## Solution

The reconciler now gates retry logic on `BurnFailureClassification`:

- **Classified failures** (`!= Unclassified`) are skipped entirely before any budget bookkeeping. A warning is emitted and the redemption stays in `Failed` until an operator drives recovery via `ResumeBurn`.
- **Orchestrator burns with a submitted transaction** are deferred rather than run through the vault-direct confirm path. The orchestrator recovery path (`OrchestratorBurnRecovered`) is the correct mechanism and is not yet automated.

A `recovery_burn_params` helper was introduced to build the mode-correct `BurnParams` from the redemption's persisted anchor (never from live config), replacing several duplicated `BurnParams::VaultDirect { .. }` constructions that would have produced the wrong shape for orchestrator-mode redemptions.

On the confirm side, `settle_reserved_burn` is now skipped for orchestrator burns (reservation accounting does not apply), and `reserve_replacement_after_revert` is skipped for non-retryable classified failures, with an error log directing operators to intervene manually.

Three new tests cover the added behaviour:

- `reconciler_skips_classified_burn_failures_without_budget` — verifies a classified failure stays `Failed` and no recovery budget events are emitted.
- `reconciler_confirms_submitted_orchestrator_burn` — verifies a mined orchestrator burn is confirmed and never resubmitted.
- `reconciler_redrives_burn_after_orchestrator_recovers` — verifies a pre-submit health-gate deferral completes on the next reconciler pass once the orchestrator is healthy.

The `MockVaultService` gains a `set_orchestrator_readiness` mutator (replacing the builder-only pattern) to support mid-test state transitions, and the unused `with_orchestrator_burn_result` builder was removed.

## Checks

By submitting this for review, I'm confirming I've done the following:

[...]

PR #274, opened 2026-07-18
feat: add `BurnParams` enum and orchestrator burn mode support

rouzwelt:
## Motivation

The redemption pipeline previously only supported vault-direct burns, where burn parameters (`vault`, `burns`, `dust_shares`, `owner`) were flat fields on `IntendBurn` and `BurnTokens` commands. Adding orchestrator-mode burns requires a different parameter shape (`token`, `amount`, `owner`) and different on-chain interactions (`submit_orchestrator_burn`, `confirm_orchestrator_burn`), with no per-receipt plan and dust retained in the bot wallet rather than returned atomically.

## Solution

Introduces a `BurnParams` enum with `VaultDirect` and `Orchestrator` variants that replaces the flat burn fields on `IntendBurn` and `BurnTokens`. The command handlers cross-check the variant against the redemption's persisted `burn_mode` anchor and return a `BurnModeMismatch` error on mismatch, preventing a caller from driving a vault-direct redemption down the orchestrator path or vice versa.

Two new events are added to the orchestrator path:

- `OrchestratorBurnSubmitted` — the counterpart of `BurnTxSubmitted`, carrying no per-receipt plan since the orchestrator walks receipts on-chain.
- `OrchestratorTokensBurned` — terminal success carrying `shares_burned`, a `burn_range` from the on-chain `Burned` event, and `dust_retained` derived from the redemption's own persisted `AlpacaCalled.dust_quantity` rather than from the command input (SPEC Decision 6).

`RedemptionError::Vault` gains a `classification` field populated at the `VaultError` boundary via a new `vault_error_to_redemption` helper, so orchestrator revert reasons (`InsufficientReceipts`, `VaultLogicMismatch`, `ReceiptLogicMismatch`) are carried through to `RecordBurnFailure` without parsing the error message string. Previously the classification was always hardcoded to `Unclassified` at the call sites.

The view and event-sourced state machine are updated to treat `OrchestratorBurnSubmitted` as an internal detail (view stays `Burning`) and `OrchestratorTokensBurned` as a terminal success transition to `Completed`, mirroring the vault-direct equivalents. The `next_burn_retry_external_tx_id_from_history` function is extended to derive the retry `externalTxId` from `OrchestratorBurnSubmitted` the same way it does from `BurnTxSubmitted`.

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [ ] linked any relevant issues or PRs

[...]

PR #273, opened 2026-07-18
feat: orchestrator burn vault service implementation

rouzwelt:
## Motivation

The redemption pipeline needs to support burning shares via the `ST0xOrchestrator` contract, which walks on-chain receipts itself rather than requiring the bot to plan per-receipt burns. Without this, orchestrator-mode redemptions cannot be executed.

## Solution

Adds four new methods to the `VaultService` trait covering the full orchestrator burn lifecycle:

- `check_orchestrator_burn_readiness` — evaluates the ERC-20 allowance gate first, then `vaultLogicIsExpected()`, returning a typed `OrchestratorBurnReadiness` outcome. Allowance is checked before the health gate so an approval shortfall is always surfaced even when the orchestrator is halted.
- `prepare_orchestrator_burn_tx` — builds and signs the `ST0xOrchestrator.burn()` transaction without broadcasting it.
- `submit_orchestrator_burn` — broadcasts the pre-signed transaction, tolerating a node-already-holds-it error on retry.
- `confirm_orchestrator_burn` — polls for the receipt, parses the `Burned` event, and on a mined-but-reverted transaction replays it as an `eth_call` at the mined block to decode the typed revert reason (`InsufficientReceipts`, `VaultLogicMismatch`, `ReceiptLogicMismatch`, or `Unknown`) into `VaultError::OrchestratorReverted`.

The supporting types (`OrchestratorBurnParams`, `OrchestratorBurnResult`, `OrchestratorBurnReadiness`, `OrchestratorRevertReason`) are introduced in a new `vault/orchestrator.rs` module. The `MockVaultService` is extended with full implementations of all four methods and test-only builder methods (`with_orchestrator_readiness`, `with_orchestrator_confirm_revert`, `with_orchestrator_burn_result`) and call-count accessors for unit test assertions.

NOTE: `#[allow(dead_code)]` are used to pass clippy, and will be removed on upstack PRs as implementation completes

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [ ] linked any relevant issues or PRs

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end orchestrated burn support, including readiness checks, transaction preparation, submission, and confirmation.
  * Confirmation now provides burned amount, receipt range, gas usage, and block details.
  * Added clear handling for failed burns caused by insufficient receipts, vault or receipt logic mismatches, and unknown errors.
[...]

PR #271, opened 2026-07-18
Persist `burn_mode` anchor and `BurnFailureClassification` on redemption events

rouzwelt:
## Motivation

Redemptions that span an asset's `VaultMode` config cutover (e.g. from `VaultDirect` to `Orchestrator`) were at risk of silently switching burn paths mid-flight, because the mode was re-resolved from live config at each burn step rather than being anchored at detection time. Additionally, burn failures carried only a free-text error string with no typed classification, forcing downstream logic to parse strings for retry-exclusion and admin grouping decisions.

## Solution

`VaultMode` now derives `serde::Serialize` / `serde::Deserialize`, making it embeddable in persisted events. Its wire format (`"VaultDirect"` / `{"Orchestrator":{"address":"0x…"}}`) is treated as a permanent event schema contract.

A `burn_mode` field is added to `RedemptionMetadata` and to the `Detected`, `Reprocessed`, `BurnResumed` events (and their corresponding commands). At transfer detection time, `detect_transfer` resolves the asset's mode from `VaultModeConfig` once and writes it onto the `Detected` event. Every subsequent burn step reads `metadata.burn_mode` from the aggregate rather than querying config, so a config change mid-redemption has no effect on in-flight redemptions. The `TransferPoller` receives a `VaultModeConfig` at construction and passes it through to `detect_transfer`.

A `BurnFailureClassification` enum is introduced on `BurningFailed` events and the `RecordBurnFailure` command. Variants cover `Unclassified` (default for historical events), `InsufficientReceipts`, `AllowanceInsufficient`, `VaultLogicMismatch`, and `ReceiptLogicMismatch`. All current call sites emit `Unclassified`; the enum is the extension point for future typed classification. The classification is propagated through to `RedemptionView::BurnFailed`.

`VaultModeConfig` gains a `mode_for` method and a `new` constructor, and `Config::vault_mode_for` delegates to it. Backward compatibility is preserved throughout: all new event fields carry `#[serde(default)]`, so historical events without them replay as `VaultDirect` / `Unclassified`.  
  
Closes [RAI-1434](https://linear.app/makeitrain/issue/RAI-1434/redemption-burn-mode-anchoring)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

## Summary by CodeRabbit

* **New Features**
  * Vault mode can now be configured per asset, with a default fallback.
[...]

PR #300, opened 2026-07-28
feat: orchestrator mint confirmation with nonce-replay recovery and failure classification

rouzwelt:
## Motivation

Orchestrator-mode mints previously had no end-to-end submission or confirmation path. The `prepare_orchestrator_mint_tx`, `confirm_orchestrator_mint`, and `find_orchestrator_minted_log` vault methods existed but were gated behind `#[allow(dead_code)]` and never wired into the mint lifecycle. Mints anchored to `VaultMode::Orchestrator` would silently fall through to vault-direct preparation, producing incorrect audit data and bypassing the orchestrator's receipt custody model entirely.

Additionally, mint failures carried no typed classification, making it impossible to distinguish environment-wide halts (`VaultLogicMismatch`, `ReceiptLogicMismatch`) from ordinary retryable failures or nonces consumed by a different mint (`NonceConsumedByOtherMint`), all of which require different resolution paths.

## Solution

Orchestrator-mode mints now have a complete submission and confirmation path in both the command handler (`Mint::execute_orchestrator_mint_confirmation`) and the durable job (`ConfirmMintJob::confirm_orchestrator`).

**Submission:** `PrepareMint` and `SubmitMintJob` branch on the persisted `VaultMode` anchor. An orchestrator mint without its recipient authorization defers silently — no event, no vault call — leaving the mint in `Minting` until the liquidity bot delivers. Once authorized, preparation routes to `prepare_orchestrator_mint_tx` carrying the anchored orchestrator address and the delivered `MintAuthorization`.

**Confirmation:** Success emits the new `OrchestratorTokensMinted` event carrying the consumed authorization `nonce` in place of a receipt id — the orchestrator holds receipt custody, so no receipt is registered. A `NonceReplayed` revert triggers a full-match lookup of the `Minted` log on `(to, nonce, token, amount)`: a hit emits `OrchestratorMintRecovered` and completes the mint; no hit means a different mint consumed the nonce, which fails as `NonceConsumedByOtherMint` for manual reconciliation only.

**Failure classification:** `MintFailureClassification` is introduced on `MintingFailed` events and the `MintingFailed` aggregate state. Typed classifications (`VaultLogicMismatch`, `ReceiptLogicMismatch`, `NonceConsumedByOtherMint`) are never auto-retried and never kick the recovery scheduler. `Unclassified` preserves the existing automatic retry behavior. The field defaults to `Unclassified` so historic events replay correctly.

**Mode enforcement:** `RecordTokensMinted` rejects orchestrator-anchored mints, and `RecordOrchestratorTokensMinted` / `RecordOrchestratorMintRecovered` reject vault-direct-anchored mints, preventing cross-mode completions. `receipt_id` on `CallbackPending` and `Completed` is now `Option<U256>`, and a new `mint_nonce: Option<B256>` field carries the orchestrator audit data.

The `#[allow(dead_code)]` suppressions on the three orchestrator vault trait methods are removed now that all three have production callers.  
  
Closes [RAI-1617](https://linear.app/makeitrain/issue/RAI-1617/orchestrator-mint-confirmation-with-nonce-replay-recovery-and-failure)

## Checks

## Summary by CodeRabbit

* **New Features**
  * Added orchestrator-backed mint submission, confirmation, and recovery.
  * Added detailed mint failure classifications and tracking for mint nonces and transaction details.
* **Bug Fixes**
  * Prevented duplicate recovery jobs from repeated authorization requests.
  * Improved handling of replayed, mismatched, and unresolved mint confirmations.
  * Preserved signer reservations across additional mint resolution events.
* **Improvements**
  * Environment-related failures can avoid consuming retry attempts.
  * Mint status responses remain compatible with older records.

### DIFF
--- a/src/mint/cmd.rs
+++ b/src/mint/cmd.rs
@@ -2,6 +2,7 @@ use alloy::primitives::{Address, B256, U256};
 use serde::{Deserialize, Serialize};
 use uuid::Uuid;
 
+use super::event::MintFailureClassification;
 use super::{
     ClientId, IssuerMintRequestId, MintExternalTxId, Network, Quantity,
     TokenSymbol, TokenizationRequestId, UnderlyingSymbol,
@@ -94,6 +95,43 @@ pub(crate) enum MintCommand {
     RecordMintFailed {
         issuer_request_id: IssuerMintRequestId,
         error: String,
+        /// Typed cause when one was decoded; typed classifications are never
+        /// auto-retried. Deliberately NOT `#[serde(default)]` (see
+        /// `Initiate::mint_mode`): commands have no historical payloads, and
+        /// a silently defaulted `Unclassified` would re-enable automatic
+        /// retries for a deterministic failure.
+        classification: MintFailureClassification,
+    },
+
+    /// Records the outcome of a successful orchestrator mint confirmation
+    /// performed by a durable `ConfirmMintJob`. Pure: produces
+    /// `OrchestratorTokensMinted` from the payload, no I/O. Idempotent — a
+    /// no-op if the mint already advanced past `TxSubmitted`. The handler
+    /// rejects vault-direct mints — an orchestrator result can never complete
+    /// a vault-direct mint, or vice versa.
+    RecordOrchestratorTokensMinted {
+        issuer_request_id: IssuerMintRequestId,
+        /// The signing-backend transaction the report belongs to; a mismatch
+        /// against the stored submission is rejected.
+        tx_id: TxId,
+        tx_hash: B256,
+        nonce: B256,
+        shares_minted: U256,
+        gas_used: u64,
+        block_number: u64,
+    },
+
+    /// Records an already-landed orchestrator mint discovered by the
+    /// `Minted`-log lookup after a `NonceReplayed` revert, full-matched on
+    /// `(to, nonce, token, amount)`. Pure: produces
+    /// `OrchestratorMintRecovered`. Idempotent — a no-op once the mint has
+    /// advanced past `TxSubmitted`.
+    RecordOrchestratorMintRecovered {
+        issuer_request_id: IssuerMintRequestId,
+        tx_hash: B256,
+        nonce: B256,
+        shares_minted: U256,
+        block_number: u64,
     },
 
     /// Retries a failed mint by transitioning `MintingFailed` -> `Minting`,
--- a/src/mint/event.rs
+++ b/src/mint/event.rs
@@ -12,6 +12,68 @@ use super::{
     TokenizationRequestId, UnderlyingSymbol,
 };
 
+/// Typed classification of an on-chain mint failure, persisted on
+/// `MintingFailed`. Mirrors the burn side's `BurnFailureClassification`:
+/// typed classifications are never auto-retried, and the halted-orchestrator
+/// cases are surfaced distinctly from ordinary mint failures.
+#[derive(
+    Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize,
+)]
+pub(crate) enum MintFailureClassification {
+    /// No typed cause was decoded — retryable on the automatic schedule.
+    #[default]
+    Unclassified,
+    /// `vaultLogicIsExpected()` mismatch — the orchestrator is halted
+    /// pending upgrade; resolves environment-wide, never per-mint.
+    VaultLogicMismatch,
+    /// The orchestrator's receipt-logic version lock tripped — halted like
+    /// `VaultLogicMismatch`.
+    ReceiptLogicMismatch,
+    /// Assigned by recovery's full-match check (not decoded from a revert):
+    /// a `Minted` log at the `(to, nonce)` pair WAS found and its token or
+    /// amount disagrees with this mint's — affirmative proof a different
+    /// mint consumed the pair, so this mint can never land. Manual
+    /// reconciliation only.
+    NonceConsumedByOtherMint,
+    /// Assigned by recovery's full-match check (not decoded from a revert):
+    /// `nonceUsed(to, nonce)` reports the nonce consumed but NO `Minted`
+    /// log at the pair was found at all. The two cannot both be true of a
+    /// healthy chain view, so the lookup itself is untrusted (window too
+    /// narrow, RPC error, indexer lag) — an unknown outcome, never proof:
+    /// this mint may well have landed. Non-retryable for SUBMISSION (the
+    /// nonce is consumed either way) but retryable for RECONCILIATION:
+    /// recovery re-runs the log query over a widened window, and a later
+    /// full match resolves the mint forward (SPEC "Recipient Authorization"
+    /// -> "Nonce").
+    NonceReplayUnresolved,
+    /// `BadRecipientSignature()` revert — the recipient signature did not
+    /// recover to `to`. Deterministic for the stored authorization: only
+    /// `CloseMint` plus a fresh `Initiate` with a new authorization resolves
+    /// it (see SPEC "Failure States").
+    BadRecipientSignature,
+    /// `RecipientCallbackRejected(recipient)` revert — an `IMintRecipient`
+    /// contract refused the mint via its `authorizeMint` callback.
+    /// Deterministic like `BadRecipientSignature`.
+    RecipientCallbackRejected,
+    /// `VaultAmountMismatch(expected, actual)` revert — the orchestrator's
+    /// on-chain 1:1 assertion failed (e.g. a share-ratio rebase mid-flight).
+    /// Alert-and-investigate; the same aggregate resumes via admin reprocess
+    /// once the ratio is restored (see SPEC "Failure States").
+    VaultAmountMismatch,
+}
+
+impl MintFailureClassification {
+    /// Whether this failure is an orchestrator-wide halt
+    /// (`vaultLogicIsExpected()` / receipt-logic version lock). Per SPEC
+    /// ("Failure States"), a halt resolves environment-wide by upgrade,
+    /// never per-mint, so it must not advance the mint's automatic-retry
+    /// attempt counter — otherwise every in-flight mint would burn its whole
+    /// budget waiting out one halt.
+    pub(crate) const fn is_environment_halt(self) -> bool {
+        matches!(self, Self::VaultLogicMismatch | Self::ReceiptLogicMismatch)
+    }
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub(crate) enum MintEvent {
     Initiated {
@@ -66,6 +128,13 @@ pub(crate) enum MintEvent {
         issuer_request_id: IssuerMintRequestId,
         error: String,
         failed_at: DateTime<Utc>,
+        /// Typed failure classification. Absent for pre-orchestrator events,
+        /// which replay as `Unclassified`. Typed classifications are never
+        /// auto-retried: `NonceConsumedByOtherMint` needs manual
+        /// reconciliation and the logic-mismatch halts resolve
+        /// environment-wide.
+        #[serde(default)]
+        classification: MintFailureClassification,
     },
     MintCompleted {
         issuer_request_id: IssuerMintRequestId,
@@ -130,6 +199,35 @@ pub(crate) enum MintEvent {
         mint_authorization: MintAuthorization,
         received_at: DateTime<Utc>,
     },
+
+    /// Orchestrator-mode mint succeeded on-chain (the counterpart of
+    /// `TokensMinted`), parsed from the orchestrator's `Minted` event.
+    /// Carries the consumed authorization `nonce` in place of vault-direct's
+    /// `receipt_id` — the orchestrator, not the bot, holds receipt custody,
+    /// so there is no bot-side receipt to record or register.
+    OrchestratorTokensMinted {
+        issuer_request_id: IssuerMintRequestId,
+        tx_hash: B256,
+        /// `Minted.nonce` — the authorization nonce this mint consumed.
+        nonce: B256,
+        shares_minted: U256,
+        gas_used: u64,
+        block_number: u64,
+        minted_at: DateTime<Utc>,
+    },
+
+    /// An orchestrator-mode mint that already landed on-chain, discovered by
+    /// recovery's `Minted`-log lookup full-matching
+    /// `(to, nonce, token, amount)` after a `NonceReplayed` revert. Carries
+    /// no `gas_used` — a bare log does not expose it.
+    OrchestratorMintRecovered {
+        issuer_request_id: IssuerMintRequestId,
+        tx_hash: B256,
+        nonce: B256,
+        shares_minted: U256,
+        block_number: u64,
+        recovered_at: DateTime<Utc>,
+    },
 }
 
 impl DomainEvent for MintEvent {
@@ -167,6 +265,12 @@ impl DomainEvent for MintEvent {
             }
             Self::MintAuthorizationReceived { .. } => {
                 "MintEvent::MintAuthorizationReceived".to_string()
+            }
+            Self::OrchestratorTokensMinted { .. } => {
+                "MintEvent::OrchestratorTokensMinted".to_string()
+            }
+            Self::OrchestratorMintRecovered { .. } => {
+                "MintEvent::OrchestratorMintRecovered".to_string()
             }
         }
     }
--- a/src/mint/job.rs
+++ b/src/mint/job.rs
@@ -28,8 +28,9 @@ use tracing::{debug, error, info, warn};
 
 use super::recovery::{enqueue_scheduled_mint_recovery, release_terminal_job};
 use super::{
-    IssuerMintRequestId, Mint, MintCommand, Quantity, TokenizationRequestId,
-    UnderlyingSymbol, has_unresolved_signer_intent,
+    IssuerMintRequestId, Mint, MintCommand, MintFailureClassification,
+    Quantity, TokenizationRequestId, UnderlyingSymbol,
+    has_unresolved_signer_intent, orchestrator_mint_failure_classification,
 };
 use crate::alpaca::{AlpacaError, AlpacaService, MintCallbackRequest};
 use crate::burn_excess::has_unresolved_excess_burn_intent;
@@ -40,8 +41,10 @@ use crate::receipt_inventory::{
 };
 use crate::tokenized_asset::Network;
 use crate::vault::{
-    MintTxStatus, NetworkVaultServices, PreparedMintTx, ReceiptInformation,
-    TxId, UnconfiguredNetworkError, VaultError, VaultService,
+    MintAuthorization, MintTxStatus, MintedLogQuery, MintedLogScan,
+    NetworkVaultServices, OrchestratorMintParams, OrchestratorRevertReason,
+    PreparedMintTx, ReceiptInformation, TxId, UnconfiguredNetworkError,
+    VaultError, VaultService,
 };
 
 /// Failure of a mint side-effect job. A domain rejection is recorded as a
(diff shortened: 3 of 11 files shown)


rouzwelt:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/300?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#300** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/300?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/300?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=300).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/mint/job.rs
```diff
@@ -643,6 +713,218 @@ impl Job<ConfirmMintContext> for ConfirmMintJob {
...
+                MintCommand::RecordMintFailed {
+                    issuer_request_id: self.issuer_request_id.clone(),
+                    error,
+                    classification,
+                },
+            )
+            .await?;
+
+        if classification == MintFailureClassification::Unclassified {
+            kick_mint_recovery(
```
0xgleb: should fix: as of this diff the claim that typed classifications are never auto-retried is enforced only by skipping the immediate `kick_mint_recovery` here. Nothing downstream consults `classification`. The reconciler runs every 300s, `find_all_recoverable_mints` matches `MintingFailed { .. }` unconditionally, and `automatic_retry_decision` reads only `failed_at` and `attempts`, so it returns `Ready` after the ordinary backoff and re-submits. The burn side this mirrors gates the recovery driver itself on the classification; here there is no equivalent, so a `NonceConsumedByOtherMint` mint is resubmitted with the same consumed nonce and reverts on every attempt. The new test only asserts the confirm job enqueued no recovery job, which documents the skipped kick rather than proving no retry.

Stack note: #301 closes the automatic path, adding `AutomaticRetryDecision::ManualOnly` so `automatic_retry_decision` short-circuits on any typed classification before budget bookkeeping, and filtering classified failures out of `find_all_recoverable_mints`. So this is an intermediate-state gap. Worth knowing that #301 replaces it with the opposite problem rather than a clean fix: with the reconciler no longer scanning these mints and the admin re-drive routed through the same short-circuited function, a classified failure ends up with no working recovery path at all. Raised on that PR.

src/mint/mod.rs
```diff
@@ -1469,12 +1660,14 @@ impl Mint {
...
             failed_at: existing_failed_at,
+            classification: existing_classification,
             attempts,
             ..
         } = self
         {
             *existing_error = error;
             *existing_failed_at = failed_at;
+            *existing_classification = classification;
             *attempts += 1;
```
0xgleb: should fix: `attempts` is incremented unconditionally here, but SPEC.md states the opposite as an explicit exception, that a `VaultLogicMismatch` or `ReceiptLogicMismatch` classified failure does not advance the counter, precisely because the halt is environment-wide and resolves by upgrade rather than per-mint. Combined with the retry gap above, every mint in flight during an orchestrator halt burns its whole automatic budget in about an hour and forty minutes, and once exhausted `manual_recovery_decision` returns `Unrecoverable` so admin reprocess rejects it too. These are post-journal mints, so the shares are already journaled to the custodian but never tokenized. Leave `attempts` unchanged for the two halt classifications.

src/mint/job.rs
```diff
@@ -188,17 +197,55 @@ impl Job<SubmitMintContext> for SubmitMintJob {
...
+                            // recorded: the mint stays in `Minting`, visible
+                            // in `/admin/stuck` past the threshold, and the
+                            // next drive retries once the liquidity bot
+                            // delivers.
+                            let Some(authorization) = mint_authorization else {
+                                drop(wallet_guard);
+                                warn!(
+                                    target: "mint",
+                                    issuer_request_id = %self.issuer_request_id,
+                                    "Orchestrator mint is awaiting its \
```
0xgleb, line 662: should fix: an orchestrator mint with no authorization defers silently here, no event and no vault call, leaving it in `Minting`. I could not find what wakes it when the authorization finally arrives. The authorization endpoint records the event on the aggregate but does not appear to re-enqueue a submit job, and the recovery reconciler only picks up `MintingFailed`, not a `Minting` mint that deferred. If that is right, a mint that arrives before its authorization stays parked indefinitely. Worth either enqueueing the submit from the authorization path or widening the reconciler to cover deferred mints.

src/mint/job.rs
```diff
@@ -643,6 +713,218 @@ impl Job<ConfirmMintContext> for ConfirmMintJob {
...
+                    error.to_string(),
+                    MintFailureClassification::Unclassified,
+                )
+                .await?;
+                return Ok(());
+            }
+        };
+
+        match vault_service
+            .find_orchestrator_minted_log(
```
0xgleb: should fix: on a `NonceReplayed` revert this completes the mint from a `Minted` log matched on `(to, nonce, token, amount)`, but that tuple does not identify which mint produced the log. Two mints to the same recipient for the same token and amount that collide on a nonce would match each other's log, so this can complete mint A from mint B's on-chain event and record shares that were never minted for A. The full-match is strictly better than a bare `(to, nonce)` hit, but it is still not a proof of ownership. Binding the lookup to something mint-specific, or refusing to auto-complete when more than one candidate mint could match the tuple, would close it.

src/mint/mod.rs
```diff
@@ -435,6 +456,54 @@ struct ConfirmedMint {
...
+/// Only the environment-wide logic-version halts classify from a revert;
+/// `NonceConsumedByOtherMint` is assigned exclusively by the full-match
+/// check after a `NonceReplayed` revert, and everything else stays
+/// `Unclassified` (retryable).
+const fn orchestrator_mint_failure_classification(
+    error: &VaultError,
+) -> MintFailureClassification {
+    match error {
+        VaultError::OrchestratorReverted { reason, .. } => match reason {
+            OrchestratorRevertReason::VaultLogicMismatch => {
```
0xgleb, line 502: should fix: several deterministic revert reasons still map to `Unclassified` and therefore keep the ordinary retry behaviour. A revert that will fail identically on every attempt should not consume the retry budget, which is the whole point of introducing the classification. Worth walking the revert reasons and typing the ones that cannot succeed on a retry.

src/mint/job.rs
```diff
@@ -643,6 +713,218 @@ impl Job<ConfirmMintContext> for ConfirmMintJob {
...
+            .confirm_orchestrator_mint(&self.tx_id)
+            .await
+        {
+            Ok(result) => {
+                info!(
+                    target: "mint",
+                    issuer_request_id = %self.issuer_request_id,
+                    tx_hash = %result.tx_hash,
+                    nonce = %result.nonce,
+                    shares_minted = %result.shares_minted,
```
0xgleb: should fix: on success this records the amount and nonce as reported by the chain event without checking them against the authorized `(amount, nonce)` the mint was prepared with. Those should be identical, and if they ever diverge that is exactly the case worth failing loudly on rather than persisting the chain value as the audit record. An equality check before emitting `OrchestratorTokensMinted` costs nothing and turns a silent divergence into a typed error.

src/mint/cmd.rs
```diff
@@ -93,6 +94,41 @@ pub(crate) enum MintCommand {
...
+        gas_used: u64,
+        block_number: u64,
+    },
+
+    /// Records an already-landed orchestrator mint discovered by the
+    /// `Minted`-log lookup after a `NonceReplayed` revert, full-matched on
+    /// `(to, nonce, token, amount)`. Pure: produces
+    /// `OrchestratorMintRecovered`. Idempotent — a no-op once the mint has
+    /// advanced past `TxSubmitted`.
+    RecordOrchestratorMintRecovered {
```
0xgleb, line 129: should fix: the mode enforcement and idempotency on `RecordOrchestratorMintRecovered` are not exercised through the command handler. Given this is the path that completes a mint from a recovered on-chain log, and the guard is what stops a cross-mode completion, it deserves a test that drives the actual command rather than only the apply function.

src/mint/mod.rs
```diff
@@ -1039,6 +1119,112 @@ impl Mint {
...
+
+                Ok(vec![MintEvent::OrchestratorMintRecovered {
+                    issuer_request_id,
+                    tx_hash,
+                    nonce,
+                    shares_minted,
+                    block_number,
+                    recovered_at: Utc::now(),
+                }])
+            }
```
0xgleb, line 1329: minor: `RecordExistingMint` looks like the one completion path that did not get a mode guard, while `RecordTokensMinted` and both orchestrator commands did. If that is deliberate because the path cannot be reached for an orchestrator-anchored mint, a comment saying so would stop the next reader wondering; if not, it is the remaining hole in the cross-mode enforcement this PR adds.

src/mint/job.rs
```diff
@@ -1476,6 +1763,385 @@ mod tests {
...
+        let test = "confirm_mint_job_nonce_consumed_by_other_mint_never_kicks";
+        assert!(logs_contain_at!(
+            Level::ERROR,
+            &[test, "consumed by a different mint"]
+        ));
+    }
+
+    /// Typed logic-mismatch halts are recorded without a recovery kick (they
+    /// resolve environment-wide), while an unclassified revert still kicks
+    /// the automatic retry.
```
0xgleb, line 5042: minor: `confirm_mint_job_kicks_recovery_only_for_unclassified` drives a path that emits a `warn!` and never asserts on it, and the same is true of several of the other new job tests. The warn is the operator-visible signal that a mint was classified non-retryable, which is the behaviour this PR is adding, so it is worth asserting rather than leaving the capture unused.

src/mint/job.rs
```diff
@@ -643,6 +713,218 @@ impl Job<ConfirmMintContext> for ConfirmMintJob {
 }
 
 impl ConfirmMintJob {
+    /// Orchestrator-mode confirmation: the job counterpart of
```
0xgleb: minor: this doc comment points at a function that does not exist under that name, so the reference dead-ends for the next reader. Worth pointing it at the real one.

src/mint/mod.rs
```diff
@@ -4733,4 +5178,288 @@ pub(crate) mod tests {
...
+            AutomaticRetryDecision::Exhausted
+        );
+
+        // The external_tx_id attempt never advanced — retries reuse retry-1.
+        assert_eq!(mint.next_retry_attempt(), 1);
+    }
+
+    /// An orchestrator mint whose authorization has not arrived must defer —
+    /// A vault-direct confirmation result can never complete an orchestrator
+    /// mint, and vice versa — both record handlers enforce the persisted
```
0xgleb: minor: the docstring on `record_handlers_reject_cross_mode_results` describes what a different test does, so it misdescribes the test it sits on. Easy to fix now and confusing later.

src/mint/mod.rs
```diff
@@ -4733,4 +5178,288 @@ pub(crate) mod tests {
...
+    /// permanent `event_type` names.
+    #[test]
+    fn orchestrator_mint_events_round_trip() {
+        let issuer_request_id = IssuerMintRequestId::random();
+        let minted = MintEvent::OrchestratorTokensMinted {
+            issuer_request_id: issuer_request_id.clone(),
+            tx_hash: B256::ZERO,
+            nonce: test_mint_authorization().nonce,
+            shares_minted: uint!(100_000000000000000000_U256),
+            gas_used: 21_000,
```
0xgleb, line 7190: minor: `orchestrator_mint_events_round_trip` mostly asserts that serde round-trips a derived impl, which is testing the derive rather than any domain behaviour. The version of this worth keeping is one that pins the on-disk wire shape, since these events are replayed, so asserting against a literal JSON fixture would catch a field rename that a round-trip never will.

src/mint/mod.rs
```diff
@@ -2146,6 +2516,10 @@ pub(crate) enum MintError {
         "Mint authorization cannot be accepted in state {current_state}: authorizations are only accepted before the mint transaction is prepared"
     )]
     AuthorizationNotAcceptable { current_state: String },
+    #[error(
+        "Mint result mode mismatch: the mint's persisted anchor is {expected}, the recorded result is {found}"
```
0xgleb, line 2828: nit: `MintModeMismatch` carries the two modes as `String`, so the error can be constructed with any text and the compiler cannot help. Carrying `VaultMode` values, or a small mode enum, keeps it honest and formats the same.

0xgleb (changes requested):

src/mint/job.rs
rouzwelt: A cross-mint tuple match requires two mints prepared with the same nonce: nonces are 256-bit values generated randomly per authorization by the liquidity bot (SPEC "Recipient Authorization" → "Nonce"), so an accidental collision is cryptographically negligible, and deliberate reuse is the bot violating the nonce contract — the same reasoning that scoped the nonce design in #298. On-chain, (to, nonce) can only ever be consumed once, so exactly one landing exists for the tuple; the full-match already refuses a differing token/amount as NonceConsumedByOtherMint. A candidate-mint uniqueness scan would add a cross-aggregate read model to defend against a 2^-256 event.

src/mint/job.rs
rouzwelt: Correct as an intermediate-state observation, and deliberately closed one commit up: #301 (directly upstack in this stack, merging together with it) adds AutomaticRetryDecision::ManualOnly short-circuiting on any typed classification before budget bookkeeping and filters classified failures out of the recoverable-mints query. Duplicating that gating here would just conflict with the commit that owns it. The "no working recovery path at all" concern you raised on #301 was addressed there — the admin path stays functional.

src/mint/event.rs
```diff
@@ -11,6 +11,68 @@ use super::{
...
+    /// Whether this failure is an orchestrator-wide halt
+    /// (`vaultLogicIsExpected()` / receipt-logic version lock). Per SPEC
+    /// ("Failure States"), a halt resolves environment-wide by upgrade,
+    /// never per-mint, so it must not advance the mint's automatic-retry
+    /// attempt counter — otherwise every in-flight mint would burn its whole
+    /// budget waiting out one halt.
+    pub(crate) const fn is_environment_halt(self) -> bool {
+        matches!(self, Self::VaultLogicMismatch | Self::ReceiptLogicMismatch)
+    }
+}
```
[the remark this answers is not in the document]
rouzwelt: Correct as an intermediate-state observation, and deliberately closed one commit up: #301 (directly upstack, merging together with this PR) adds AutomaticRetryDecision::ManualOnly/ReconcileOnly so the decision short-circuits on typed classifications before budget bookkeeping, and filters classified failures out of the recoverable-mints query. Duplicating that gating here would conflict with the commit that owns it.

src/mint/mod.rs
```diff
@@ -333,7 +334,15 @@ pub(crate) enum Mint {
...
-        receipt_id: U256,
+        /// Vault-direct audit data: the ERC-1155 receipt id the deposit
+        /// minted to the bot. `None` for orchestrator-mode mints — the
+        /// orchestrator holds receipt custody, and `mint_nonce` is the
+        /// analogous proof.
+        receipt_id: Option<U256>,
+        /// Orchestrator audit data: the authorization nonce the mint
+        /// consumed. `None` for vault-direct mints.
+        #[serde(default)]
+        mint_nonce: Option<B256>,
```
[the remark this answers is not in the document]
rouzwelt: The suggestion breaks a compatibility contract the state deliberately maintains: these are persisted snapshot/view payloads, and pre-orchestrator rows carry a bare receipt_id (see the mint_mode doc on the same variant — "Snapshots persisted before orchestrator mode default to VaultDirect"). A minted_proof enum can't #[serde(default)] its way to compatibility, so it would require custom deserialization shims on both Mint and MintView for historic financial rows. mint_mode is already the persisted discriminant, both fields are set at exactly two apply sites (each writing a coherent pair), and apply is the only constructor — so the invalid combinations have no producer. The trade is a real serde-migration risk against a type-level impossibility that nothing can currently reach; happy to revisit once the orchestrator stack lands and the compat window closes.

src/mint/mod.rs
```diff
@@ -1058,6 +1159,140 @@ impl Mint {
...
+                    return Err(MintError::MintedNonceMismatch {
+                        expected: authorization.nonce,
+                        actual: nonce,
+                    });
+                }
+                let authorized_shares = quantity
+                    .to_u256_with_18_decimals()
+                    .map_err(|error| MintError::QuantityConversion {
+                        message: error.to_string(),
+                    })?;
```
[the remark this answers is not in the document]
rouzwelt: #[from] isn't available here: MintError derives Clone, PartialEq, Serialize, Deserialize (src/mint/mod.rs:2594), so embedding QuantityConversionError forces those derives onto it — and its ParseFailed(#[from] rust_decimal::Error) variant can't satisfy Serialize/Deserialize (rust_decimal's error isn't serde-serializable). Restructuring the shared conversion type to allow it would be a cross-cutting change beyond this PR; the message shape mirrors RedemptionError::QuantityConversion (src/redemption/mod.rs:1575), which exists for exactly the same reason.

src/mint/mod.rs
```diff
@@ -365,6 +374,10 @@ pub(crate) enum Mint {
         journal_confirmed_at: DateTime<Utc>,
         error: String,
         failed_at: DateTime<Utc>,
+        /// Typed failure cause from `MintingFailed.classification`; typed
+        /// classifications are never auto-retried.
```
0xgleb, line 380: should fix: Typed failure classifications are persisted and the confirmation job skips the immediate kick, but `automatic_retry_decision` and `drive_one_step` still treat every `MintingFailed` as submission-retryable. A classified environment halt can therefore re-enter `RetryMint` without advancing attempts and repeatedly resubmit the same mint. Gate non-`Unclassified` failures out of automatic retry and add a recovery regression test, as the burn path already does.

src/mint/event.rs
```diff
@@ -11,6 +11,68 @@ use super::{
...
+    /// `nonceUsed(to, nonce)` reports the nonce consumed but NO `Minted`
+    /// log at the pair was found at all. The two cannot both be true of a
+    /// healthy chain view, so the lookup itself is untrusted (window too
+    /// narrow, RPC error, indexer lag) — an unknown outcome, never proof:
+    /// this mint may well have landed. Non-retryable for SUBMISSION (the
+    /// nonce is consumed either way) but retryable for RECONCILIATION:
+    /// recovery re-runs the log query over a widened window, and a later
+    /// full match resolves the mint forward (SPEC "Recipient Authorization"
+    /// -> "Nonce").
+    NonceReplayUnresolved,
```
0xgleb, line 48: should fix: NonceReplayUnresolved is documented as reconciliation-only (widened Minted-log lookup), but confirm always uses lookback_blocks: None, recovery has no re-query path, and RecordOrchestratorMintRecovered rejects MintingFailed. What actually happens later is RetryMint + resubmit of a consumed nonce. Please add a real reconciling drive that re-scans with a widened window and can complete from the parked failure without preparing a new mint tx.

0xgleb (changes requested):

src/mint/event.rs
rouzwelt: fixed in #301

src/mint/mod.rs
rouzwelt: fixed in #301

0xgleb (approved):

JuaniRios (approved):

PR #301, opened 2026-07-28
feat: orchestrator mint recovery with classified failure handling

rouzwelt:
## Motivation

Orchestrator mints follow a different custody model than vault-direct mints: the orchestrator holds custody of the on-chain receipt rather than the bot, so the existing vault-direct recovery paths (receipt lookup, `RecoverFromReceipt`) cannot be used for them. Without orchestrator-aware recovery, a failed orchestrator mint has no safe path to completion or closure — resubmitting blindly risks double-minting, and closing blindly risks stranding real backed tokens unreported to Alpaca.

Additionally, certain failure classifications (e.g. `NonceConsumedByOtherMint`, `VaultLogicMismatch`) require manual operator intervention and must never be automatically retried, but the automatic retry scheduler had no way to express this distinction.

## Solution

**Orchestrator recovery paths:**

- `TxIntended` and `Minting` recovery now branch on `VaultMode`: vault-direct uses the existing receipt lookup; orchestrator mints use a new `recover_from_landed_orchestrator_mint` helper that queries for a landed `Minted` log full-matching `(to, nonce, token, amount)` as the double-mint guard before any resubmission.
- A new `confirm_recovery_orchestrator_mint` method handles the known-tx confirmation branch for orchestrator mints. A `NonceReplayed` revert here — after the landed-log pre-check found nothing — means a different mint consumed the nonce, which produces a `NonceConsumedByOtherMint` classified failure requiring manual reconciliation rather than a resubmission. Any other confirmed revert consumed nothing, so a fresh transaction is submitted. Transient errors defer to the next pass.
- The existing `confirm_recovery_vault_direct_mint` logic is extracted into its own method to keep the two paths cleanly separated.
- `RecoverFromReceipt` now explicitly rejects orchestrator mints with `MintModeMismatch`, since the bot never custodies a receipt for them and a receipt-discovery delivery would be a mis-attribution.

**Classified failure handling:**

`AutomaticRetryDecision::ManualOnly(classification)` stops the scheduler and reconciler on any typed classification before budget bookkeeping; halt classifications don't advance the attempt counter (per SPEC); the admin reprocess endpoint pushes a `manual`\-flagged recovery job that spends exactly one re-drive of a classified failure, so "recovery initiated" always means something ran. `RecordExistingMint` rejects orchestrator-anchored mints (`MintModeMismatch`) and the submit job treats that rejection as terminal. **Safe close for orchestrator mints:** the pre-close check gates on `nonceUsed` (unbounded non-landing proof), full-matches the bounded `Minted`\-log scan, honours the adjudicated `NonceConsumedByOtherMint` verdict, and otherwise fails closed (`MintLanded` / `LandingUnproven` 422s; lookup errors refuse with 500). Vault-direct closes are untouched. Adds the `nonce_used` vault read and the `Mint::mint_authorization` accessor.

**Safe close for orchestrator mints:**

- `handle_close_mint` is made `async` and now runs the landed-`Minted`\-log check for orchestrator mints that have received an authorization. A full-match refuses the close with `CloseRefusedMintLanded` — the operator must run recovery instead. A lookup failure also refuses the close (fail-closed). Vault-direct closes are unchanged and never run the check.

**New** **`mint_authorization`** **accessor:**

- A `mint_authorization` const accessor is added to `Mint` to allow the close handler and other callers to retrieve the delivered authorization without pattern-matching every state.

Closes [RAI-1618](https://linear.app/makeitrain/issue/RAI-1618/orchestrator-mint-recovery-with-classified-failure-handling)

## Checks

## Summary by CodeRabbit

- **New Features**
  - Added safer mint closure validation using transaction, receipt, nonce, and event checks.
  - Added nonce acknowledgement support for unresolved replay cases.
  - Added manual mint recovery options for eligible failures.

- **Bug Fixes**
  - Prevented closure when mint completion cannot be proven or verification fails.
  - Stopped futile automatic retries for deterministic authorization failures.
  - Improved recovery reconciliation for unresolved nonce replays.

- **Tests**
  - Expanded coverage for closure validation, recovery decisions, reconciliation, and failure handling.

### DIFF
--- a/src/admin.rs
+++ b/src/admin.rs
@@ -1,11 +1,13 @@
+use alloy::consensus::transaction::SignerRecoverable;
 use alloy::network::ReceiptResponse;
-use alloy::primitives::{Address, B256};
+use alloy::primitives::{Address, B256, U256};
 use apalis_sqlite::SqlitePool as ApalisSqlitePool;
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
 use cqrs_es::AggregateError;
 use event_sorcery::{EventSourced, LifecycleError, Store};
-use rocket::http::Status;
+use rocket::http::{ContentType, Status};
+use rocket::response::{self, Responder};
 use rocket::serde::json::Json;
 use rocket::{get, post};
 use serde::{Deserialize, Serialize};
@@ -21,12 +23,14 @@ use crate::auth::InternalAuth;
 use crate::config::{Config, VaultMode};
 use crate::mint::{
     IssuerMintRequestId, ManualRecoveryDecision, Mint, MintCommand, MintEvent,
-    MintView, TokenizationRequestId, find_stuck as find_stuck_mints,
+    MintFailureClassification, MintView, TokenizationRequestId,
+    find_stuck as find_stuck_mints,
     recovery::{
-        ManualRetryOutcome, enqueue_scheduled_mint_recovery,
-        manually_retry_failed_mint,
+        ManualRetryOutcome, enqueue_manual_mint_recovery,
+        enqueue_scheduled_mint_recovery, manually_retry_failed_mint,
     },
 };
+use crate::receipt_inventory::ReceiptService;
 use crate::redemption::Redemption;
 use crate::redemption::burn_manager::{
     BurnManager, BurnManagerError, MAX_AUTOMATIC_BURN_RECOVERY_ATTEMPTS,
@@ -40,10 +44,11 @@ use crate::redemption::{
     next_burn_retry_external_tx_id_from_history,
 };
 use crate::tokenized_asset::schedule::{FreezeScheduleError, FreezeScheduler};
-use crate::tokenized_asset::view::list_enabled_assets;
+use crate::tokenized_asset::view::{find_vault, list_enabled_assets};
 use crate::tokenized_asset::{Network, UnderlyingSymbol};
 use crate::vault::{
-    BurnVerification, NetworkVaultServices, TxId, VaultError, VaultService,
+    BurnTxStatus, BurnVerification, MintedLogQuery, MintedLogScan,
+    NetworkVaultServices, SendableTxWithHash, TxId, VaultError, VaultService,
 };
 
 #[async_trait]
@@ -1282,7 +1287,12 @@ const fn map_burn_manager_error(err: &BurnManagerError) -> Status {
         (status = 400, description = "Aggregate id is not a valid UUID"),
         (status = 404, description = "No mint found for this id"),
         (status = 409, description = "Mint already completed or closed"),
-        (status = 422, description = "Invalid state transition for recovery"),
+        (status = 422,
+            description = "Invalid state transition for recovery, or a \
+                deterministic recipient-authorization failure \
+                (BadRecipientSignature / RecipientCallbackRejected) that a \
+                re-drive can never fix — close the mint and re-initiate \
+                with fresh authorization"),
         (status = 500, description = "View query or internal failure")
     ),
     security(("internal_api_key" = []))
@@ -1296,7 +1306,7 @@ pub(crate) async fn reprocess_mint(
     store: &rocket::State<Arc<Store<Mint>>>,
     vault_services: &rocket::State<NetworkVaultServices>,
     aggregate_id: &str,
-) -> Result<Json<ReprocessResponse>, Status> {
+) -> Result<Json<ReprocessResponse>, ReprocessMintError> {
     let issuer_request_id: IssuerMintRequestId = aggregate_id
         .parse::<uuid::Uuid>()
         .map(IssuerMintRequestId::new)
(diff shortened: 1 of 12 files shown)


rouzwelt:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/301?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#301** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/301?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/301?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=301).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

tests/turnkey_cutover.rs
```diff
@@ -188,13 +189,53 @@ async fn mint_before_cutover(
...
+        let completed: i64 = sqlx::query_scalar(
+            "
+            SELECT COUNT(*)
+            FROM events
+            WHERE aggregate_type = 'Mint'
+              AND event_type = 'MintEvent::MintCompleted'
+            ",
+        )
+        .fetch_one(&pool)
+        .await?;
```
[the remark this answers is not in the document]
rouzwelt: There's no public endpoint exposing committed completion — a Completed mint drops out of /admin/stuck, so an API poll would assert vacuous absence. The DB poll is inter-phase synchronization (same established pattern as wait_for_settled_burn in this file), not scenario verification, which remains on the public boundary.

src/mint/recovery.rs
```diff
@@ -748,6 +748,18 @@ async fn recover_mint_until_automatic_budget_exhausted(
                     reason: AbandonReason::AutomaticRetriesExhausted,
                 };
             }
+            // Typed classifications are never auto-retried: the scheduled
+            // loop stops here, before any wakeup/budget bookkeeping, and the
+            // admin re-drive is the only way forward (mirrors the burn
+            // side's classified-skip).
+            AutomaticRetryDecision::ManualOnly(classification) => {
```
0xgleb, line 971: critical: this arm says the admin re-drive is the only way forward, and the description says the operator re-drive bypasses the classified skip, but the re-drive goes through this same function. `reprocess_mint` only calls `enqueue_scheduled_mint_recovery`, whose job calls `recover_mint_until_automatic_budget_exhausted`, which now returns `Resolved` here without driving a step. It is worse than a no-op: `manual_recovery_decision` maps only `Exhausted` to `Unrecoverable`, and a classified failure now returns `ManualOnly` rather than `Exhausted`, so it falls through to `Eligible` and the endpoint answers 200 recovery initiated while nothing runs. With classified failures also excluded from `find_all_recoverable_mints`, the mint is never re-enqueued either. The burn side works because its skip lives in `recover_single_burn_failed`, which the admin endpoint bypasses via a separate `ResumeBurn` command; the mint skip sits in the only path the mint admin endpoint has. Net effect is that `NonceConsumedByOtherMint` and `VaultLogicMismatch` mints become permanently unrecoverable, and the operator's only remaining lever is the close this PR argues is unsafe. Either plumb a force flag from `reprocess_mint` so this arm is only taken on the automatic path, or stop reporting `Eligible` for a classified failure so the endpoint at least fails honestly.

src/admin.rs
```diff
@@ -1372,6 +1385,168 @@ pub(crate) async fn close_mint(
...
+        );
+        Status::InternalServerError
+    })?;
+
+    let landed = vault_service
+        .find_orchestrator_minted_log(
+            orchestrator,
+            wallet,
+            authorization.nonce,
+            vault,
```
0xgleb: should fix: this treats `Ok(None)` from `find_orchestrator_minted_log` as proof nothing landed, but that lookup only scans `MINTED_LOG_LOOKBACK_BLOCKS`, 250k blocks or roughly 5.8 days on Base, and returns `Ok(None)` for anything older. That window was sized for the confirm and recovery timeline, where a `None` is safe because it converges on `NonceConsumedByOtherMint`. Here the same `None` lets the close proceed. An admin close is an unbounded-latency action on a parked mint, and the mints most likely to be parked longest are exactly the classified ones, so the guard fails open precisely in the case it exists for: closing a mint whose tokens really did land, stranding backed tokens unreported to Alpaca. Either widen the window for this call or treat an out-of-window answer as inconclusive and refuse, consistent with the fail-closed posture the rest of the check takes.

src/mint/mod.rs
```diff
@@ -5462,4 +5594,60 @@ pub(crate) mod tests {
...
+    }
+
+    /// Typed classifications opt out of the automatic retry schedule before
+    /// any budget bookkeeping; only `Unclassified` stays on it.
+    #[test]
+    fn automatic_retry_decision_manual_only_for_typed_classifications() {
+        for classification in [
+            MintFailureClassification::VaultLogicMismatch,
+            MintFailureClassification::ReceiptLogicMismatch,
+            MintFailureClassification::NonceConsumedByOtherMint,
```
0xgleb, line 7845: should fix: the branch this test guards is ordered so classification wins over exhaustion, but every case here builds a mint with `attempts == 1`, well under the budget of 4. So the test cannot tell always wins apart from only consulted while under budget. That distinction is behavioural, not cosmetic: `Exhausted` re-checks on-chain receipt existence and returns `Abandoned`, while `ManualOnly` returns `Resolved` with no receipt check. If someone later reorders the two `if`s, an already-retried classified failure silently takes the wrong recovery path and every assertion here still passes. Drive `attempts` past the budget in one case and assert the decision is still `ManualOnly`.

src/admin.rs
```diff
@@ -1372,6 +1385,168 @@ pub(crate) async fn close_mint(
...
+    };
+
+    let vault = find_vault(pool, underlying, &network)
+        .await
+        .map_err(|err| {
+            error!(target: "admin", issuer_request_id = %issuer_request_id,
+                error = %err,
+                "Vault lookup failed for the pre-close landed check"
+            );
+            Status::InternalServerError
```
0xgleb: should fix: failing closed on a lookup error is right, but it makes deterministic failures permanent on the operator's last-resort endpoint. A missing vault-mode config entry or a persistently failing RPC is not transient, so the close returns 500 forever with no way through, on the one action left for a mint that cannot be recovered. Worth separating the transient case, where refusing and retrying is correct, from the deterministic one, where the operator needs either a specific actionable error or an explicit acknowledged override.

src/mint/mod.rs
```diff
@@ -1322,15 +1360,36 @@ impl Mint {
...
+                issuer_request_id: expected_id,
+                mint_mode,
+                ..
+            } => {
                 Self::validate_issuer_request_id(
                     expected_id,
                     &issuer_request_id,
                 )?;
 
+                if let VaultMode::Orchestrator { .. } = mint_mode {
```
0xgleb, line 1663: minor: the new orchestrator rejection on `RecordExistingMint` is not handled by its only caller, so it surfaces as a generic retryable job error rather than the permanent domain refusal it is. It will be retried until the budget is spent, and the operator sees a retry-exhausted mint instead of a mode mismatch. Worth matching it at the call site and treating it as terminal.

src/admin.rs
```diff
@@ -1372,6 +1385,168 @@ pub(crate) async fn close_mint(
...
+    store: &Store<Mint>,
+    pool: &Pool<Sqlite>,
+    vault_services: &NetworkVaultServices,
+    issuer_request_id: &IssuerMintRequestId,
+) -> Result<(), CloseMintError> {
+    let mint = store.load(issuer_request_id).await.map_err(|err| {
+        error!(target: "admin", issuer_request_id = %issuer_request_id,
+            error = %err,
+            "Failed to load mint for the pre-close landed check"
+        );
```
0xgleb, line 1751: minor: the fail-closed branches of the pre-close check have no test coverage. The behaviour they encode, that a lookup failure refuses the close, is a deliberate safety decision, and it is the kind that quietly inverts during a refactor because nothing pins it. A test driving a lookup error and asserting the close is refused would hold it in place.

src/mint/recovery.rs
```diff
@@ -748,6 +748,18 @@ async fn recover_mint_until_automatic_budget_exhausted(
...
+                warn!(target: "mint", issuer_request_id = %issuer_request_id,
+                    classification = ?classification,
+                    "Skipping non-retryable classified mint failure; manual \
+                     recovery required"
+                );
```
0xgleb, line 1031: minor: the description credits this diff with orchestrator recovery machinery in job.rs that already existed before it. It makes the change look larger than it is and sends a reviewer looking for code that is not in the diff. Worth trimming the description to what this PR actually adds, which is the classified-failure handling and the safe close.

src/admin.rs
```diff
@@ -6279,4 +6459,321 @@ mod tests {
...
+        let mint_store = Arc::new(test_store::<Mint>(pool.clone(), ()));
+        let issuer_request_id = seed_close_test_mint(
+            &mint_store,
+            VaultMode::Orchestrator { address: CLOSE_ORCHESTRATOR },
+        )
+        .await;
+
+        let vault_mock = Arc::new(MockVaultService::new_success());
+        let (status, _body) = dispatch_close_mint(
+            close_mint_rocket(&pool, mint_store.clone(), vault_mock.clone()),
```
0xgleb, line 7350: minor: the new close-mint tests assert the returned behaviour but not the logs, though they exercise paths that emit them and the project asks for log assertions. For a refused close in particular the log is the operator-facing signal explaining why, so it is worth pinning.

0xgleb (changes requested):

src/admin.rs
rouzwelt: Partly resolved by the nonceUsed-first rework: the vault lookup and quantity conversion now run only after the nonce is known consumed, so a deterministic conversion/lookup failure can no longer block the close of a mint that provably never landed (unconsumed nonce ⇒ immediate allow). One correction on the premise: the orchestrator address comes from the mint's persisted mint_mode anchor, not from vault-mode config, so the "missing config entry" case doesn't exist on this path. What remains 500-forever is: no vault service for the network (deployment config — fixable), and a consumed nonce with a persistently failing lookup — where the guard genuinely cannot distinguish "landed long ago" from "not ours", and closing on an acknowledged override is precisely the strand-backed-tokens outcome the check exists to prevent. The refusal logs name the failing step; the remedy is operational. Keeping fail-closed without an override here is deliberate.

src/mint/recovery.rs
```diff
@@ -882,6 +1016,16 @@ async fn drive_one_step(
             let vault = resolve_vault(ctx, underlying, *network).await?;
             enqueue_submit(ctx, issuer_request_id, vault).await?;
```
0xgleb, line 1227: should fix: Minting/TxIntended still go straight to enqueue_submit with no orchestrator Minted-log FullMatch first. The PR text and SPEC treat that scan as the primary double-mint guard (NonceReplayed as fallback), and RecordOrchestratorMintRecovered was widened for these states, but nothing calls it on the live path. Please add the proactive landed-log recover before any resubmit.

src/mint/recovery.rs
```diff
@@ -748,6 +822,66 @@ async fn recover_mint_until_automatic_budget_exhausted(
...
+                    manual_redrive_available = false;
+                    info!(target: "mint", issuer_request_id = %issuer_request_id,
+                        classification = ?classification,
+                        "Operator-requested re-drive of a classified mint \
+                         failure"
+                    );
+                    if let Err(error) =
+                        drive_one_step(ctx, &mint, issuer_request_id).await
+                    {
+                        debug!(target: "mint", issuer_request_id = %issuer_request_id,
```
0xgleb, line 1018: should fix: manual redrive always calls drive_one_step, and for classified MintingFailed that means RetryMint + submit. For NonceConsumedByOtherMint / BadRecipientSignature that cannot succeed and only burns a wallet nonce; SPEC wants close + fresh auth. Please branch the manual flag by classification (re-scan vs resubmit vs refuse).

src/mint/job.rs
```diff
@@ -403,7 +405,30 @@ impl SubmitMintJob {
...
+                error!(
+                    target: "mint",
+                    issuer_request_id = %self.issuer_request_id,
+                    tx_hash = %receipt.tx_hash,
+                    receipt_id = %receipt.receipt_id,
+                    "Vault receipt mis-attributed to an orchestrator-anchored \
+                     mint; refusing vault-direct completion — manual \
+                     investigation required"
+                );
+                return Ok(true);
```
0xgleb: should fix: MintModeMismatch returns Ok(true) without changing the aggregate, so submit exits "success" while the mint stays Minting and recovery will spin until no-progress. Please record a classified/terminal failure (or equivalent park) before stopping the job.

src/mint/mod.rs
```diff
@@ -1260,7 +1314,20 @@ impl Mint {
...
+            }
+            | Self::TxSubmitted {
+                issuer_request_id: expected_id,
+                mint_mode,
+                ..
+            }
+            | Self::MintingFailed {
                 issuer_request_id: expected_id,
                 mint_mode,
                 ..
```
0xgleb, line 1384: should fix: RecordOrchestratorMintRecovered is now accepted from Minting/TxIntended/MintingFailed but still skips the nonce/shares checks that OrchestratorTokensMinted enforces. Please apply the same invariants before emitting the recovered event.

0xgleb (changes requested):

src/mint/recovery.rs
rouzwelt: fixed in #303

0xgleb (approved):

JuaniRios (approved):

PR #302, opened 2026-07-28
Add `VaultModeTag` field to `TokenizedAssetStatusResponse`

rouzwelt:
## Motivation

The liquidity bot needs to know which minting path each asset uses — either depositing directly into the vault (`VaultDirect`) or routing through the ST0xOrchestrator (`Orchestrator`) — so it can determine which assets require a signed `MintAuthV1` to be delivered before their mints can be submitted. Previously, the `GET /tokenized-assets/<underlying>/status` response had no way to communicate this, leaving the liquidity bot without a reliable signal.

## Solution

A new `VaultModeTag` enum (`vault_direct` | `orchestrator`) is introduced in the DTO crate and added as a `vault_mode` field on `TokenizedAssetStatusResponse`. The field is additive and defaults to `VaultDirect` when absent, so responses from servers that predate the field remain valid — a server that predates the field can only ever mint vault-direct, making the default truthful.

The status endpoint now reads the per-asset vault mode from the server's `Config` and maps it to the tag via a `From<VaultMode> for VaultModeTag` conversion. The orchestrator address is intentionally dropped in this mapping; the liquidity bot only needs the tag, and the issuance bot's config remains the single source of truth for addresses during the cutover.

`VaultModeTag` is also registered in the OpenAPI schema and TypeScript bindings exports, with wire format enforced as `snake_case` (matching `TokenizedAssetStatus`).

Closes [RAI-1619](https://linear.app/makeitrain/issue/RAI-1619/vaultmodetag)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

## Summary by CodeRabbit

- **New Features**
  - Tokenized asset status responses now indicate the vault mode: `vault_direct` or `orchestrator`.
  - API documentation and generated type definitions include the new vault mode information.

- **Bug Fixes**
  - Older responses without vault mode information remain compatible and default to direct vault mode.
  - Status reporting now accurately reflects mixed vault-mode configurations across assets.

PR #268, opened 2026-07-16
Add `VaultMode` config with orchestrator support and smoke test

rouzwelt:
## Motivation

The issuance service needs to support routing mint/burn operations through the `ST0xOrchestrator` contract (which handles EIP-712 mint-auth and receipt-walk for burns) as an alternative to calling `OffchainAssetReceiptVault` directly. During the pilot phase, only specific assets will use the orchestrator path while others remain on the existing vault-direct path. A safe, zero-config default (all assets vault-direct) is required so the dark-deploy state is unchanged without an explicit config file.

## Solution

Introduces a `VaultMode` enum (`VaultDirect` | `Orchestrator { address }`) and a `VaultModeConfig` struct that holds a per-asset override map and a global default. `Config::vault_mode_for(&UnderlyingSymbol)` resolves the correct mode for any given asset at call time.

Configuration is loaded from an optional TOML file passed via `--config` / `CONFIG`. The file schema (`[orchestrator]`, `[assets.<SYMBOL>]`) uses `deny_unknown_fields` so typos are startup errors rather than silent no-ops. Validation enforces that any asset or default resolving to `Orchestrator` mode requires a valid `[orchestrator].address`; missing or malformed addresses are hard startup errors.

A `config.example.toml` is committed alongside the code to document the schema and is pinned to the strict parser in a test so it can never silently drift out of sync.

The `LocalEvm` test harness gains a `deploy_orchestrator()` helper that reproduces the pinned production vault beacon set at its canonical CREATE2 addresses (via the Zoltu deterministic factory etched onto Anvil), deploys `ST0xOrchestrator` behind a beacon proxy, and grants `MINT_ROLE`/`BURN_ROLE` to the deployer wallet. A new `orchestrator_smoke` integration test exercises a full deploy → EIP-712 sign → mint → approve → burn round-trip on Anvil, asserting the share balance returns to zero.

Closes [RAI-1409](https://linear.app/makeitrain/issue/RAI-1409/impl-toml-config-and-smok-tests)  
Closes [RAI-1217](https://linear.app/makeitrain/issue/RAI-1217/2-orchestrator-bindings-config-and-anvil-test-harness)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

## Summary by CodeRabbit

* **New Features**
  * Added configurable vault execution modes, including per-asset direct or orchestrator-driven routing.
[...]

PR #267, opened 2026-07-16
Add st0x-deploy ABIs via soldeer-based Nix builder

rouzwelt:
## Motivation

The project needs ABI definitions from the `st0x-deploy` repository to interact with ST0x contracts (orchestrator, beacon, proxy, receipt vault, etc.). Previously, only `ethgild` ABIs were available, and there was no mechanism to handle Soldeer-based dependency management in the Nix build system.

## Solution

A two-phase `mkSoldeerAbi` builder was introduced to support Forge projects that use Soldeer for dependency management. The first phase is a fixed-output derivation with network access that runs `forge soldeer install` and caches the resulting `dependencies/` directory. The second phase is a sandboxed build that injects the vendored dependencies and runs `forge build` offline. This separation ensures the dependency hash only changes when `soldeer.lock` changes, not when contracts change.

The `st0x-deploy` repository is added as a flake input and wired through `abis.nix` using a new `nix/st0x-deploy.nix` module. The `abis` symlink structure is updated so that each source's compiled output lives under its own named subdirectory (`abis/ethgild` and `abis/st0x.deploy`) rather than a single flat `abis` symlink. ABI environment variables for the ST0x orchestrator, beacon, proxy, receipt, receipt vault, and beacon set deployer contracts are exposed via `abiEnv`.

Closes [RAI-1408](https://linear.app/makeitrain/issue/RAI-1408/update-nix-setup-to-build-st0xdeploy-abis)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [ ] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

## Summary by CodeRabbit

* **New Features**
  * Added support for St0x deployment artifacts and contract interfaces.
  * Added ABI generation and integration for orchestrator, beacon, proxy, receipt, and vault contracts.
  * Included new bindings for interacting with St0x orchestrator contracts.
* **Build Improvements**
  * Added offline-compatible dependency preparation for reproducible contract builds.
  * Development environments now expose the new St0x deployment artifacts alongside existing ABI files.

PR #303, opened 2026-07-28
feat: orchestrator-mode mint path with recipient authorization and e2e tests

rouzwelt:
## Motivation

Orchestrator-mode mints require a recipient authorization — the liquidity bot signs the orchestrator's EIP-712 `MintAuthV1` digest over `(token, to, amount, nonce)` and delivers it before the mint can proceed. Previously, the wire shape and correlation key for this internal call were left as implementation details to be resolved later. The `vault_mode` field was also absent from the tokenized asset status endpoint, leaving the liquidity bot without a way to know which assets require an authorization.

## Solution

The internal authorization endpoint is now settled as `POST /internal/mints/<tokenization_request_id>/authorization` with body `{ "nonce": <bytes32 hex>, "signature": <hex bytes> }`, correlated by `tokenization_request_id`. The endpoint validates the EIP-712 signer and queries `nonceUsed()` on-chain before recording the authorization. Response semantics are fully specified: `200` (idempotent re-delivery), `404` (unknown mint), `409` (conflicting authorization already recorded or mint already past intent), `422` (vault-direct asset, bad signer, or nonce already consumed), and `502` (retryable on-chain read failure).

The consumed nonce is the mint's idempotency key for recovery: startup recovery full-matches the orchestrator's `Minted` log against `(to, nonce, token, amount)` and completes without resubmitting. A nonce consumed by a different mint parks the mint as a classified failure for manual reconciliation rather than falsely completing it. The `vault_mode` field is now included in `GET /tokenized-assets/<underlying>/status` so the liquidity bot can determine which assets require an authorization.

End-to-end tests on Anvil cover the full orchestrator mint flow, kill-and-restart recovery of a mint that already landed on-chain (asserting exactly one `Minted` log and no resubmission), nonce collision parking as a `MintingFailed` stuck entry with no Alpaca callback, and a mixed-mode deployment where an orchestrator-mode asset and a vault-direct asset each take their own path concurrently.

Closes [RAI-1620](https://linear.app/makeitrain/issue/RAI-1620/orchestrator-mode-mint-path-with-recipient-authorization-and-e2e-tests)

Closes [RAI-1220](https://linear.app/makeitrain/issue/RAI-1220/5-mint-via-orchestratormint-with-recipient-authorization)

## Checks

## Summary by CodeRabbit

- **New Features**
  - Added per-asset vault-mode selection for direct and orchestrated minting.
  - Added authorization submission for orchestrated mints with nonce-based replay protection and idempotent handling.
  - Added recovery for previously completed or interrupted orchestrated mints.

- **Bug Fixes**
  - Prevented duplicate mint submissions after successful transactions.
  - Improved handling of nonce conflicts and temporary authorization lookup failures.

- **Documentation**
  - Documented vault modes, authorization flows, response outcomes, and recovery scenarios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,8 +209,38 @@ endpoints.
 2. We validate and respond with `issuer_request_id`
 3. Alpaca journals shares from AP to our custodian account
 4. Alpaca confirms journal → we receive `/inkind/issuance/confirm`
-5. We mint tokens on-chain via `vault.deposit()`
+5. We mint tokens on-chain (vault-direct: `vault.deposit()`; orchestrator:
+   `ST0xOrchestrator.mint()` — see below)
 6. We call Alpaca's callback endpoint
+
+Each mint anchors its mode at initiation from the per-asset `vault_mode` config
+(see SPEC.md "Orchestrator Migration"): vault-direct assets mint via the
+`vault.deposit()` multicall with the receipt held by the bot; orchestrator
+assets mint via a single `ST0xOrchestrator.mint()` — the orchestrator custodies
+the receipt and forwards the shares to the recipient wallet.
+
+An orchestrator-mode mint additionally requires a **recipient authorization**:
+the liquidity bot signs the orchestrator's EIP-712 `MintAuthV1` digest over
+`(token, to, amount, nonce)` with the recipient wallet's key and delivers
+`{nonce, signature}` via the internal
+`POST /internal/mints/<tokenization_request_id>/authorization` endpoint, which
+validates the signer and nonce on-chain before recording it. Until the
+authorization arrives the mint waits — it never falls back to vault-direct. The
+consumed nonce is single-use on-chain, making it the mint's idempotency key:
+recovery of a landed mint full-matches the orchestrator's `Minted` log against
+`(to, nonce, token, amount)` and completes without resubmitting. A consumed
+nonce that does not full-match takes one of two verdicts, never conflated: a log
+at the pair disagreeing on token/amount is proof a different mint consumed it
+(parked as `NonceConsumedByOtherMint` for manual reconciliation), while an empty
+scan is an inconclusive chain view (parked as `NonceReplayUnresolved`,
+re-queried by recovery over a widened window and resolved forward once the
+landing becomes visible — closable only with an explicit operator
+acknowledgement). The `vault_mode` field on
+`GET /tokenized-assets/<underlying>/status` tells the liquidity bot which assets
+will need authorizations for _new_ mints, but the authoritative requirement for
+any given mint is its own persisted `mint_mode` anchor: a mint initiated before
+a config flip keeps the mode it was initiated with, so the bot must answer the
+mint it was asked about, not the asset's current config.
 
 ## Redemption Flow
 
--- a/src/mint/job.rs
+++ b/src/mint/job.rs
@@ -153,9 +153,27 @@ impl Job<SubmitMintContext> for SubmitMintJob {
             // rebroadcast only when StillMineable/MinedSuccess; terminal
             // dead/revert records MintingFailed so recovery may replace;
             // uncertain observation preserves MintIntended (no MintingFailed).
-            Mint::TxIntended { prepared_tx, network, .. } => {
-                self.submit_from_tx_intended(ctx, prepared_tx, *network)
-                    .await?;
+            Mint::TxIntended {
+                prepared_tx,
+                network,
+                quantity,
+                wallet,
+                mint_mode,
+                mint_authorization,
+                ..
+            } => {
+                self.submit_from_tx_intended(
+                    ctx,
+                    prepared_tx,
+                    SubmitFromTxIntendedParams {
+                        network: *network,
+                        wallet: *wallet,
+                        quantity,
+                        mint_mode,
+                        mint_authorization: mint_authorization.as_ref(),
+                    },
+                )
+                .await?;
             }
             // A re-run after the submission was already recorded: the
             // confirm job may not have been enqueued before a crash, so
@@ -202,6 +220,34 @@ struct ResolvePreparedParams<'a> {
     mint_authorization: Option<&'a MintAuthorization>,
 }
 
+struct SubmitFromTxIntendedParams<'a> {
+    network: Network,
+    wallet: Address,
+    quantity: &'a Quantity,
+    /// Mode anchor from the aggregate — decides whether the pre-submit
+    /// double-mint guard runs before the prepared bytes are rebroadcast.
+    mint_mode: &'a VaultMode,
+    mint_authorization: Option<&'a MintAuthorization>,
+}
+
+/// Outcome of the orchestrator pre-submit double-mint guard
+/// ([`SubmitMintJob::recover_landed_orchestrator_mint`]).
+enum OrchestratorPreSubmitOutcome {
+    /// A landed mint full-matched and was recorded; the confirm job owns the
+    /// rest of the chain.
+    Recovered,
+    /// An on-chain read failed — nothing is provable, so submission defers
+    /// to the next recovery pass rather than proceeding blind.
+    Defer,
+    /// The nonce is consumed and a classified `MintingFailed` was recorded
+    /// (`NonceConsumedByOtherMint` on the proven mismatch,
+    /// `NonceReplayUnresolved` on an empty scan). Submission must not run;
+    /// recovery's reconciliation or the operator owns the mint now.
+    Parked,
+    /// Nothing landed under this mint's nonce; submission proceeds.
+    Proceed,
+}
+
 impl SubmitMintJob {
     /// Refuses while another persisted intent holds this signer's nonce domain.
     ///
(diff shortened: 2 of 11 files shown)


rouzwelt:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/303?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#303** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/303?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/303?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=303).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

src/mint/job.rs
```diff
@@ -168,6 +168,102 @@ impl Job<SubmitMintContext> for SubmitMintJob {
...
+                                // arm pushes the callback job, completing
+                                // the chain.
+                                self.enqueue_confirm(
+                                    ctx,
+                                    TxId::Hash(minted.tx_hash),
+                                )
+                                .await?;
+                                return Ok(());
+                            }
+                            Ok(None) => {}
```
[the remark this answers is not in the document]
rouzwelt: `Ok(None)` isn't a lookup failure — it's evidence the nonce was consumed by a different mint, and submitting is what routes that case through the confirm-side recheck into the classified `NonceConsumedByOtherMint` park. Deferring would never converge (a foreign log can never full-match), leaving the mint re-deferring forever. The lag case is covered by the confirm-side recheck running after the revert. Added a comment, a WARN, and a pinning test at the arm

rouzwelt:
@coderabbit review

rouzwelt:
@coderabbit review

src/mint/job.rs
```diff
@@ -368,7 +189,338 @@ impl Job<SubmitMintContext> for SubmitMintJob {
...
+            // not yet visible to this scan.)
+            Ok(None) => {
+                warn!(
+                    target: "mint",
+                    issuer_request_id = %self.issuer_request_id,
+                    nonce = %authorization.nonce,
+                    "Nonce is consumed but no landed mint full-matches; \
+                     submitting so the confirm-side recheck can classify"
+                );
+                Ok(OrchestratorPreSubmitOutcome::Proceed)
```
0xgleb: should fix: this arm returns `Proceed` on the reasoning that the guaranteed `NonceReplayed` revert routes through the confirm job into the classified `NonceConsumedByOtherMint` park. That route is unreachable in production. Proceeding calls `prepare_orchestrator_mint_tx`, which calls `provider.fill(tx)`, and the signing provider is built `.disable_recommended_fillers().with_gas_estimation()`, so `fill` issues `eth_estimateGas`, which reverts for a consumed nonce. Preparation returns `Err`, the failure is recorded as `Unclassified`, recovery re-drives it, and estimation fails again until the budget is spent. The mint never reaches `TxSubmitted`, so the confirm-side recheck that emits `NonceConsumedByOtherMint` never runs. The diff contains its own evidence: `send_reverting_orchestrator_mint` has to pass `.gas(1_000_000)` with a comment that estimation would refuse to broadcast, and the e2e collision test seeds `MintTxSubmitted` directly rather than exercising this path. Record the classified failure here instead of proceeding.

tests/orchestrator_mint.rs
```diff
@@ -0,0 +1,802 @@
...
+        &evm,
+        orchestrator_address,
+        &user_signer,
+        tokens(50),
+        nonce,
+    )
+    .await?;
+    deliver_mint_authorization(
+        &client,
+        tokenization_request_id,
```
0xgleb, line 363: should fix: every orchestrator scenario delivers the authorization before `confirm_mint_journal`, so the submit job always finds it populated on first drive. Production ordering is not ours to choose, since Alpaca confirms the journal on its own schedule, so the mint routinely reaches `Minting` first and the submit job takes the defer branch. Nothing in the authorize endpoint enqueues or kicks anything, so resumption depends entirely on the background recovery job. The one path only an e2e test can prove, multi-component resumption after a deferral, is the path with no e2e coverage, in the PR whose stated purpose is e2e coverage. The existing unit test asserts the mint stops, never that it starts again. Worth a case that confirms the journal first, then delivers the authorization, and asserts the mint still completes with exactly one `Minted` log and the Alpaca callback fired.

tests/orchestrator_mint.rs
```diff
@@ -0,0 +1,802 @@
...
+        &evm,
+        orchestrator_vault_modes("AAPL", orchestrator_address),
+    )?;
+    let rocket = initialize_rocket(config).await?;
+    let client = rocket::local::asynchronous::Client::tracked(rocket).await?;
+
+    let entry = wait_for_minting_failed_entry(&client).await?;
+    assert_eq!(entry["aggregate_id"], issuer_request_id.as_str());
+    assert!(
+        entry["detail"]
```
0xgleb: should fix: the collision test asserts the mint parks as a stuck `MintingFailed` with no Alpaca callback, but it cannot tell `NonceConsumedByOtherMint` from `Unclassified`, and those two mean opposite things. `Unclassified` is auto-retried; the classified one is the operator signal that must never be. So the test passes either way and the never-auto-retried claim stays unverified. Assert the classification itself, and ideally that a recovery pass leaves it alone.

src/mint/mod.rs
```diff
@@ -1226,7 +1228,20 @@ impl Mint {
         block_number: u64,
     ) -> Result<Vec<MintEvent>, MintError> {
         match self {
-            Self::TxSubmitted {
+            Self::Minting {
+                issuer_request_id: expected_id, mint_mode, ..
+            }
+            | Self::TxIntended {
+                issuer_request_id: expected_id,
```
0xgleb, line 1372: should fix: this handler widened from accepting only `TxSubmitted` to `Minting | TxIntended | TxSubmitted | MintingFailed`, but the doc comment still names two states, and the two new arms have neither a caller nor a test. Widening the states a completion command accepts is exactly where a cross-mode or double-completion bug hides, so arms that nothing exercises are worth either wiring or dropping until they have a caller. At minimum the doc comment should match what the code accepts.

tests/orchestrator_mint.rs
```diff
@@ -0,0 +1,802 @@
...
+    // the Alpaca callback is the completion signal.
+    harness::wait_for_mock_hits(&mint_callback_mock, 1).await?;
+
+    let reader = bot_provider(&evm).await?;
+    let orchestrator =
+        IST0xOrchestratorV1Instance::new(orchestrator_address, &reader);
+    let minted_logs =
+        orchestrator.Minted_filter().from_block(0).query().await?;
+    assert_eq!(
+        minted_logs.len(),
```
0xgleb, line 758: minor: the headline assertion of the recovery test, exactly one `Minted` log, cannot fail as written. It is the property the test exists to prove, that restart does not resubmit, so it should be able to fail if a second mint were submitted. Worth checking it counts logs from the chain over the full window rather than a value the harness already constrained to one.

src/vault/mock.rs
```diff
@@ -1845,6 +1886,39 @@ impl VaultService for MockVaultService {
...
+        _orchestrator: Address,
+        _to: Address,
+        nonce: B256,
+    ) -> Result<bool, VaultError> {
+        #[cfg(test)]
+        {
+            if *self.orchestrator.nonce_used_should_error.lock().unwrap() {
+                return Err(VaultError::InvalidReceipt);
+            }
+            let value = *self.orchestrator.nonce_used.lock().unwrap();
```
0xgleb, line 2009: minor: the mock ignores the `to`, `token` and `orchestrator` arguments and matches on the rest, so no unit test can catch a lookup that passes the wrong recipient or the wrong orchestrator. Since the production lookup full-matches four fields precisely because a partial match is not proof, the mock is weaker than the invariant it stands in for. Tightening it costs little and makes the mode and recipient tests meaningful.

src/mint/job.rs
```diff
@@ -368,7 +189,338 @@ impl Job<SubmitMintContext> for SubmitMintJob {
...
+            )
+            .await
+        {
+            Ok(Some(minted)) => {
+                info!(
+                    target: "mint",
+                    issuer_request_id = %self.issuer_request_id,
+                    tx_hash = %minted.tx_hash,
+                    nonce = %minted.nonce,
+                    "Found landed orchestrator mint, recording recovery \
```
0xgleb, line 1074: minor: worth a second look at whether this guard can attribute another mint's landing to this one. It matches a landed `Minted` log to decide whether to skip resubmission, and if the match can be satisfied by a different mint that shares the tuple, the outcome is completing a mint that never minted. I am not certain the collision is reachable given how the nonce is chosen, so treat this as a question rather than a defect: what guarantees the matched log belongs to this mint and not a sibling?

README.md
```diff
@@ -209,9 +209,31 @@ endpoints.
...
+Each mint anchors its mode at initiation from the per-asset `vault_mode` config
+(see SPEC.md "Orchestrator Migration"): vault-direct assets mint via the
+`vault.deposit()` multicall with the receipt held by the bot; orchestrator
+assets mint via a single `ST0xOrchestrator.mint()` — the orchestrator custodies
+the receipt and forwards the shares to the recipient wallet.
+
+An orchestrator-mode mint additionally requires a **recipient authorization**:
+the liquidity bot signs the orchestrator's EIP-712 `MintAuthV1` digest over
+`(token, to, amount, nonce)` with the recipient wallet's key and delivers
+`{nonce, signature}` via the internal
```
0xgleb, line 225: minor: this points the liquidity bot at the live-config status endpoint to decide whether an asset needs an authorization, but the requirement is anchored per mint on the persisted `mint_mode`, not on live config. A mint initiated before a config flip carries the old anchor, so a bot reading live config can reach the opposite conclusion from the mint it is answering. Worth describing the anchor as the source of truth.

src/mint/job.rs
```diff
@@ -368,7 +189,338 @@ impl Job<SubmitMintContext> for SubmitMintJob {
...
+        if let (
+            VaultMode::Orchestrator { address: orchestrator },
+            Some(authorization),
+        ) = (mint_mode, mint_authorization)
+        {
+            match self
+                .recover_landed_orchestrator_mint(
+                    ctx,
+                    &vault,
+                    *orchestrator,
```
0xgleb, line 646: minor: the double-mint guard reads the chain before the wallet lock is acquired, so the answer can age between the check and the send. The window is probably small in practice and the on-chain nonce is the real backstop, but on a guard whose whole job is preventing a double mint it is worth either moving the read inside the lock or noting why the race is benign.

src/vault/mod.rs
```diff
@@ -306,6 +306,20 @@ pub(crate) trait VaultService: Send + Sync {
         Err(VaultError::InvalidReceipt)
     }
 
+    /// Reads the orchestrator's `nonceUsed(to, nonce)` consumed flag — the
+    /// single cheap read gating [`Self::find_orchestrator_minted_log`]'s
+    /// full lookback scan: `false` proves nothing can have landed under this
+    /// mint's nonce (the on-chain uniqueness key), so the scan is skipped on
+    /// every ordinary first submission.
+    async fn nonce_used(
```
0xgleb, line 329: nit: the new trait method defaults to returning an error. For a method on the orchestrator authorization path, an implementation that forgets to override it now fails at runtime rather than at compile time, and the failure reads as an on-chain problem rather than a missing implementation. A default that cannot be silently inherited would be safer.

0xgleb (changes requested):

src/vault/mod.rs
rouzwelt: The Err(InvalidReceipt) default is the established pattern across the entire orchestrator surface of this trait (ten-plus methods, introduced downstack) — nonce_used following it keeps the surface uniform. Both real implementations override it, a forgotten override fails loudly with a typed error on the first call rather than silently succeeding, and every path through it is covered by tests that would catch that immediately. Making this one method required while its siblings default would be inconsistent; converting the whole surface to required methods is a refactor that belongs to the commit that introduced the pattern, if wanted.

src/mint/job.rs
rouzwelt: The guarantee is the nonce contract: nonces are 256-bit values generated randomly per authorization by the liquidity bot (SPEC "Recipient Authorization" → "Nonce"), and (to, nonce) is consumed at most once on-chain — so the only way a different mint satisfies the full tuple is a nonce collision between independently drawn 256-bit values (cryptographically negligible) that also matches token and amount, or deliberate reuse, which is the bot violating the contract. Within the contract, a full-tuple match is unique to this mint's signed authorization.

src/mint/job.rs
```diff
@@ -462,49 +269,457 @@ impl SubmitMintJob {
...
+        // `NonceReplayed` and reconciles through this same full-match on
+        // the next drive. This check exists to discover recoveries and
+        // spare doomed submissions, not to be the safety boundary.
+        if let (
+            VaultMode::Orchestrator { address: orchestrator },
+            Some(authorization),
+        ) = (mint_mode, mint_authorization)
+        {
+            match self
+                .recover_landed_orchestrator_mint(
```
0xgleb, line 643: should fix: The new recover_landed_orchestrator_mint guard only runs for Minting. The TxIntended arm still goes straight to submit_mint after the vault-direct receipt check, so a crash after MintTxIntended with a consumed auth nonce resubmits and lands in Unclassified via record_submission_failure instead of recover-forward / typed park. Please run the same full-match guard before rebroadcasting prepared orchestrator bytes, and cover that crash window in e2e (today’s recovery tests only seed Minting).

tests/orchestrator_mint.rs
```diff
@@ -0,0 +1,1097 @@
...
+        .header(rocket::http::ContentType::JSON)
+        .header(rocket::http::Header::new("X-API-KEY", TEST_API_KEY))
+        .remote(
+            "127.0.0.1:8000".parse().expect("test client address must parse"),
+        )
+        .body(json!({ "nonce": nonce, "signature": signature }).to_string())
+        .dispatch()
+        .await;
+
+    assert_eq!(
```
0xgleb: nit: The new orchestrator mint e2e file only drives successful authorization delivery (deliver_mint_authorization asserts 200). Wrong-signer / wrong-recipient and vault-direct rejection are covered in authorize unit tests, but not on the Anvil HTTP path this PR adds. A single 422 case that then asserts no Minted log would lock the recipient-authz boundary the liquidity bot relies on.

0xgleb (changes requested):

JuaniRios (approved):

0xgleb (approved):

PR #323, opened 2026-08-07
feat: resolve the infra age identity from 1Password via SSH_IDENTITY_OP

JuaniRios:
## Motivation

- The infra commands resolve one identity and use it for two jobs: `rage -d` on the encrypted droplet IP, and `ssh -i`. `rage` cannot use the 1Password SSH agent, cause the agent protocol only signs, and age decryption needs the raw key for key agreement.
- `SSH_IDENTITY` only accepts a file path. An operator whose keys live only in 1Password has no working bare invocation, only an explicit `--op` per command.
- Every `op read` costs an interactive 1Password approval. So the fix must not route every command through `op read` on every invocation, or the tooling becomes a prompt farm.
- This broke silently at #313: the rotation re-keyed the age files, and anyone with a stale `~/.ssh/id_ed25519` got `No matching keys found` with no hint. Closes [RAI-1683](https://linear.app/makeitrain/issue/RAI-1683/resolve-the-infra-toolings-age-identity-from-1password-via-ssh).

## Solution

The design rule: a 1Password key is for decryption only, and the only thing cached on disk is the decrypted droplet IP. Key material is never cached, never passed to `ssh -i` (SSH goes through the 1Password SSH agent), and only exists as a `0600` temp file that an EXIT trap deletes.

- `SSH_IDENTITY_OP` (+ `SSH_IDENTITY_OP_ACCOUNT`) joins the precedence chain: `--op` > `-i` > `$SSH_IDENTITY` > `$SSH_IDENTITY_OP` > `~/.ssh/id_ed25519`. Recommended home is the gitignored `.envrc.local`, sourced by `.envrc`.
- New `infra/op-identity.nix`: the single `op read` path shared by bash (`parseIdentity`) and nushell (`scripts/secret.nu`). It validates the fetched key's header, cause 1Password serves SSH-key items as PKCS#8 by default and `rage` cannot parse that. The error tells you to append `?ssh-format=openssh`.
- Resolution is lazy for op sources: `parseIdentity` records the source, and `_require_identity` materializes the key at the first decrypt. A command that never decrypts never calls `op` and never prompts.
- Host-IP cache: `{env}Remote`, `{env}Service*`, `{env}DbReset`, and the deploy scripts cache the decrypted IP (`0600` in a `0700` runtime dir), keyed by the sha256 of `infra/.remote-{env}.age`. `tf-apply`/`tf-destroy` rewrite that file, so rotation is an exact, automatic cache miss. A warm cache means zero `op` calls and `ssh` without `-i`. Cache failures warn and degrade to uncached, they never abort a command.
- `bootstrap` is the one exception to the ssh rule: `nixos-anywhere` and its install probes need a concrete key path, so it does pass the materialized key as `-i`. The docs say so.
- Fixed on the way: `rage -d ... > file` truncated the target before decrypting, so a failed decrypt in `tf-apply` re-encrypted a zero-byte `terraform.tfstate` over `terraform.tfstate.age`. Decryption now goes to a temp file and renames only on success.
- `deploy.nix` no longer clobbers the resolved identity, and the `DEPLOY_HOST` branch now consumes identity flags instead of forwarding them to `deploy` as argv.
- `secret` resolves the key in its bash wrapper so a bash EXIT trap owns the removal. Nushell cannot trap SIGINT, so Ctrl-C in `$EDITOR` used to leak the temp key.

No committed tests: this is bash generated from Nix, and the repo has no shell test harness. Verified by `nix build` of every infra/deploy/flake attr (shellcheck runs inside `writeShellApplication`) plus stub-`op` smoke runs of the generated scripts: cache hit/miss, precedence order, flag guards, signal cleanup, PKCS#8 rejection, and both deploy branches. The real `op` was never invoked.

## Checks

- [ ] added comprehensive test coverage for any changes in logic (no shell harness in the repo, see the note above)

## Summary by CodeRabbit

- **New Features**
  - Added streamlined SSH identity resolution using command-line options, environment variables, default keys, and 1Password references.
  - Added a helper for securely retrieving supported identities from 1Password.
  - Added safer host-resolution caching and temporary key handling.

- **Bug Fixes**
  - Improved deployment behavior when a host is supplied directly.
  - Added validation and clearer warnings for invalid or unavailable identities.
  - Improved cleanup and failure handling during secret editing and rekeying.

- **Documentation**
  - Updated provisioning and deployment guidance to reflect the unified identity workflow.

### DIFF
--- a/docs/nixos-provisioning.md
+++ b/docs/nixos-provisioning.md
@@ -20,14 +20,87 @@ are **kebab-case**.
 
 ```bash
 # from repo root — either:
-nix run .#tfEditVars -- -i "$SSH_IDENTITY"
+nix run .#tfEditVars
 # or:
 nix develop
-tf-edit-vars -i "$SSH_IDENTITY"
+tf-edit-vars
 ```
 
-Identity: `export SSH_IDENTITY=~/.ssh/id_ed25519`, or pass `-i`, or
-`--op 'op://vault/item'`.
+### Identity resolution
+
+`tf*`, `{env}Remote`, `{env}Service*`, `bootstrap`, `rekey`, `secret`, and the
+deploy scripts all resolve one decryption identity through the same precedence:
+`--op <uri>` (+ `--op-account <acct>`, accepted only alongside `--op`) >
+`-i <key>` > `$SSH_IDENTITY` > `$SSH_IDENTITY_OP` (+ optional
+`$SSH_IDENTITY_OP_ACCOUNT`) > `~/.ssh/id_ed25519`. Every command in this doc is
+shown bare — pass one of the flags, as the first argument, only to override the
+default. With an `op://` source the key file is materialized for decryption only
+and is never passed to `ssh -i`, so SSH itself authenticates through the
+1Password SSH agent. `bootstrap` is the exception: it provisions a fresh
+droplet, and both `nixos-anywhere` and its pre/post-install SSH probes need a
+concrete key path, so bootstrap does pass the materialized key as `ssh -i` /
+`IdentityFile=` for the lifetime of that command.
+
+For every command but `bootstrap`, SSH access therefore requires the 1Password
+SSH agent to be serving a key that is present in the server's `authorized_keys`
+— the `op://` item used for decryption does not have to be that SSH key, and
+cannot be when it holds an age identity.
+
+An SSH agent cannot substitute for any of this: age decryption needs the raw
+private key for key agreement with the recipient, and the SSH agent protocol
+exposes only a sign operation, no way to derive a shared secret from the key it
+holds. `DEPLOY_HOST=<ip>` deploys are the one exception (see "Deploy" below) —
+they skip decryption entirely, so `SSH_IDENTITY_OP` has no effect there and an
+`op://` source leaves the 1Password SSH agent as the only credential. A file
+identity (`-i <key>` or `$SSH_IDENTITY`) is still passed to ssh as `-i`.
+
+**Recommended setup:** put your 1Password identity in the gitignored
+`.envrc.local` (sourced by `.envrc`; run `direnv allow` once after creating or
+editing it):
+
+```bash
+echo "export SSH_IDENTITY_OP='op://<vault>/<item>/private key'" >> .envrc.local
+# optional, only if the item is not in your default 1Password account:
+echo "export SSH_IDENTITY_OP_ACCOUNT=<account>" >> .envrc.local
+direnv allow
+```
+
+The bare field URI above is right for an item that stores the key in a **text
+field**. For a 1Password **SSH key** item, append `?ssh-format=openssh` —
+without it `op read` serves the key as PKCS#8, which `rage` cannot parse (see
+the
+[`op read` reference](https://developer.1password.com/docs/cli/reference/commands/read)).
+The tooling checks the fetched key's format and says so if it is wrong.
+
+**Caching:** no key material is ever cached. Every `--op`/`SSH_IDENTITY_OP`
+resolution is a fresh `op read` into a `0600` temp file that is deleted when the
+command exits, Ctrl-C included: every command resolves the key in its shell
+wrapper, where an EXIT trap owns the removal. That resolution is also lazy — it
+only happens the moment a command actually needs to decrypt something, so a
+command that never decrypts never calls `op` and is never prompted. What IS
+cached is the **decrypted droplet IP** (`{env}Remote`, `{env}Service*`,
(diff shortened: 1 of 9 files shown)


JuaniRios:
* **#323** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/323?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/323?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=323).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

infra/default.nix
```diff
@@ -24,15 +30,19 @@ let
...
+          # decryption can never leave a truncated file at the real path --
+          # otherwise a zero-byte terraform.tfstate here would get
+          # re-encrypted over terraform.tfstate.age by the EXIT trap.
+          _decrypt_tmp=$(mktemp ${file}.XXXXXX)
+          _decrypt_tmpfiles+=("$_decrypt_tmp")
+          if rage -d -i "$identity" ${file}.age > "$_decrypt_tmp"; then
+            mv "$_decrypt_tmp" ${file}
+          else
+            rm -f "$_decrypt_tmp"
             exit 1
```
0xgleb, line 44: should fix: A failing decrypt now leaves any pre-existing `terraform.tfstate` untouched. Because `preambleWithEncrypt` has already installed its EXIT handler, this `exit 1` still runs `syncRemotes` and `state.encrypt`, so stale plaintext can overwrite the valid ciphertext. Remove or quarantine the plaintext before decrypting, or suppress synchronization and encryption when decryption fails.

infra/op-identity.nix
```diff
@@ -0,0 +1,114 @@
...
+    fi
+
+    tmp=$(mktemp)
+    # Not an EXIT trap: the file is the deliverable, so the trap only covers the
+    # window before the caller learns the path and takes over deletion.
+    trap 'on_signal INT' INT
+    trap 'on_signal TERM' TERM
+    trap 'on_signal HUP' HUP
+    fetch_key "$uri" "$account" "$tmp"
+    trap - INT TERM HUP
```
0xgleb, line 111: should fix: The signal traps are installed only after `mktemp` and removed before the path is printed, leaving handoff windows where neither this helper nor the caller can clean the temporary key. A signal in either window can leave the private key on disk. Make the caller provide the destination path, or use another explicit ownership handoff that has no unowned interval.

infra/default.nix
```diff
@@ -169,20 +274,65 @@ let
...
+      resolveHostBody = ''
+        ${hostIpCacheLib}
+
+        _host_ip_dir=""
+        _host_ip_entry=""
+        host_ip=""
+        if _host_ip_dir=$(_ensure_host_ip_cache_dir) && [ -f ${remoteFile.agePath} ]; then
+          # The hash names the entry, so a failure to read the ciphertext leaves
+          # $_host_ip_entry empty: no read, no write-back, command unaffected.
+          if _host_ip_age_hash=$(sha256sum ${remoteFile.agePath} 2>/dev/null | cut -d' ' -f1) &&
```
0xgleb, line 288: should fix: The cache key is computed from the ciphertext, but the miss path decrypts that file later without locking or rechecking its hash. A concurrent Terraform rewrite can publish the new host under the old hash, while a cache hit can act on the old host during rotation. Recheck the hash after decryption or serialize hashing, decryption, and cache publication.

docs/nixos-provisioning.md
```diff
@@ -20,14 +20,87 @@ are **kebab-case**.
...
+nix run .#tfEditVars
 # or:
 nix develop
-tf-edit-vars -i "$SSH_IDENTITY"
+tf-edit-vars
 ```
 
-Identity: `export SSH_IDENTITY=~/.ssh/id_ed25519`, or pass `-i`, or
-`--op 'op://vault/item'`.
+### Identity resolution
```
0xgleb, line 29: should fix: These identity-precedence, temporary-key, cache, and deployment contracts are defined only in the runbook, and the PR adds no committed regression coverage for them. The repository workflow requires new behavior to be specified in `SPEC.md` and covered by durable tests. Add the security invariants to the specification and tests for failure cleanup, signal cleanup, cache hit and miss behavior, and `DEPLOY_HOST` routing.

0xgleb (changes requested):

PR #317, opened 2026-08-06
Move `MintAuthorizationRequest`/`Response` DTOs to shared `st0x-issuance-dto` crate

rouzwelt:
## Motivation

The `MintAuthorizationRequest` and `MintAuthorizationResponse` types were defined locally inside the issuance bot's `authorize.rs` handler, making them inaccessible to the liquidity bot's client crate. The client needed these types to deliver signed `MintAuthV1` recipient authorizations to `POST /internal/mints/<tokenization_request_id>/authorization`, but had no way to reference them without duplicating the definitions.

## Solution

`MintAuthorizationRequest` and `MintAuthorizationResponse` are promoted from the internal `authorize.rs` module into `st0x_issuance_dto`, becoming the shared wire types for both sides of the service-to-service channel. The server-side handler now imports them from the DTO crate and converts `IssuerMintRequestId` to `String` when constructing responses to match the now-generic `issuer_request_id: String` field. The OpenAPI schema registration is updated to reference the DTO crate types directly.

`IssuanceClient` gains a `deliver_mint_authorization` method that `POST`s a `MintAuthorizationRequest` to the authorization endpoint, using the same path-segment URL construction as the existing status endpoint to preserve base path prefixes and percent-encode path-significant characters in the tokenization request ID. Every non-`200` response is surfaced as `ClientError::Status` — including `404`, which is a retryable race condition (mint not yet initiated) rather than a benign "not found" that should collapse to `None`.

Tests cover the happy path with exact hex wire format assertions, empty-signature delivery (`"0x"`), all four contract status codes (`404`, `409`, `422`, `502`) surfacing as errors, malformed response body handling, and prefixed/encoded URL construction.  
  
Issuance part of [RAI-1243](https://linear.app/makeitrain/issue/RAI-1243/liquidity-bot-sign-and-deliver-mintauthv1-for-orchestrator-mode-mints)  
Closes [RAI-1673](https://linear.app/makeitrain/issue/RAI-1673/mint-authorization-request-for-dto-and-client-issuance-side)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
- [x] made this PR as small as possible
- [x] linked any relevant issues or PRs

## Summary by CodeRabbit

* **New Features**
  * Added support for submitting signed mint authorization data.
  * Added structured mint authorization responses, including request identifiers and authorization status.
  * Improved support for encoded authorization endpoints and empty signatures.

* **Bug Fixes**
[...]

PR #309, opened 2026-07-31
Add orchestrator onboarding CLI subcommands, SPEC docs, and runbook

rouzwelt:
## Motivation

The orchestrator rollout requires a defined, verifiable onboarding procedure before any asset's `vault_mode` can flip to `"orchestrator"`. Previously, there was no documented runbook or CLI tooling to establish and confirm the preconditions (role grants, Turnkey signing policy, per-asset ERC-20 approvals) in a way that gates the cutover deterministically.

## Solution

Three new `issuer` subcommands are introduced and documented in `SPEC.md`:

- `issuer orchestrator-preflight` — read-only gate that verifies `MINT_ROLE` + `BURN_ROLE` on the orchestrator for the Turnkey bot wallet, checks `vaultLogicIsExpected()`, and confirms each asset's unlimited ERC-20 approval to the orchestrator is in place. Exits non-zero unless every check passes, so runbook steps can gate on it.
- `issuer approve-orchestrator <UNDERLYING>` — executes the one-time unlimited ERC-20 approval (bot wallet → orchestrator) through the Turnkey signer. Idempotent against an already-unlimited allowance, and success is confirmed by an on-chain allowance read rather than inferred from the receipt.
- `issuer verify-orchestrator-signing <UNDERLYING>` — signs one transaction per required shape (`mint`, `burn`, `approve`, ERC-1155 `safeTransferFrom`) without broadcasting, surfacing any Turnkey policy denial by name before the pilot's first live mint.

All three subcommands require the `TURNKEY_*` env group and derive the orchestrator address exclusively from the TOML config file — no address arguments are accepted.

A new runbook (`docs/runbooks/orchestrator-onboarding.md`) provides the ordered procedure: shipping the orchestrator address dark in config, coordinating role grants with st0x.deploy, configuring and proving the Turnkey signing policy, executing per-asset approvals, and running the final preflight gate. It includes a record table for prod facts (orchestrator address, role grant transactions, `EMERGENCY_ROLE` holder, policy names, per-asset approval hashes) and an escalation path for `InsufficientReceipts` and `VaultLogicMismatch` failure states.

The `SPEC.md` prose for the existing `force-complete-redemption` section was also reflowed for readability with no semantic changes.

Closes [RAI-1221](https://linear.app/makeitrain/issue/RAI-1221/6-ops-turnkey-policy-role-grants-approvals)

## Checks

By submitting this for review, I'm confirming I've done the following:

- [x] added comprehensive test coverage for any changes in logic
[...]

PR #308, opened 2026-07-31
Add orchestrator onboarding CLI subcommands (preflight, approve, verify-signing)

rouzwelt:
## Motivation

Before an asset's `vault_mode` can flip to orchestrator, the ops runbook requires several pre-cutover checks: the Turnkey bot wallet must hold `MINT_ROLE` and `BURN_ROLE` on the orchestrator, the orchestrator's `vaultLogicIsExpected()` must report healthy, and each asset must have its one-time unlimited ERC-20 approval (bot → orchestrator on the vault share token) executed. Previously there was no tooling to verify or execute these steps before cutover, meaning gaps would only surface during a live customer flow.

Additionally, the orchestrator address was not retained in `VaultModeConfig` when no asset had yet flipped to orchestrator mode, making it unavailable to onboarding tooling that needs it before the first cutover.

## Solution

`VaultModeConfig` now stores the `[orchestrator].address` from the TOML file independently of whether any asset resolves to orchestrator mode, exposed via `orchestrator_address()`. The file-loading logic is extracted into a public `load_vault_mode_config` function shared by server startup and the new CLI subcommands.

Three new issuer CLI subcommands are added:

- **`orchestrator-preflight`**: Read-only gate that checks roles, `vaultLogicIsExpected()`, and per-asset allowances on-chain, printing a structured report and exiting non-zero unless every check passes. The orchestrator address is always sourced from the TOML config file, never typed.
- **`approve-orchestrator`**: Executes the one-time unlimited ERC-20 approval for a single asset through the Turnkey signer. Idempotent — an already-unlimited allowance sends nothing. Requires an explicit confirmation naming the asset, vault, orchestrator, and wallet before any network call.
- **`verify-orchestrator-signing`**: Signs (without broadcasting) one transaction per shape the Turnkey signing policy must allow before cutover: `mint` and `burn` on the orchestrator, `approve` on the vault share token, and ERC-1155 `safeTransferFrom` on the receipt contract. A policy gap surfaces here as a named `SigningRejected` error rather than during a live mint.

The underlying on-chain logic lives in a new `vault/onboarding.rs` module (`check_orchestrator_readiness`, `ensure_unlimited_approval`, `prove_signing_shapes`), generic over the provider and independently testable. A `test_wallet_against` helper is added to `wallet/turnkey.rs` to allow tests outside that module to simulate Turnkey signing-policy denials via a mock server.

## Checks

By submitting this for review, I'm confirming I've done the following:

[...]

PR #306, opened 2026-07-29
docs: restore overwritten mint and freeze spec

0xgleb:
## Stack Context

A later custody-migration restack retained an older whole-file version of `SPEC.md`, overwriting the durable mint-job documentation from #208 and the scheduled freeze-window documentation from #239 while leaving both implementations on `main`.

The parent PR #307 stabilizes the custody-rehearsal test race exposed while validating this documentation repair.

## What?

- restore the durable mint command, recovery, sequence, and state-machine specification
- restore the scheduled freeze-window API and job semantics
- correct restored command fields/transitions against the current implementation
- update the stale mint-flow summary in `AGENTS.md`
- preserve the custody-migration and legacy force-completion documentation that landed afterward

## Why?

`SPEC.md` is the project's source of truth, but it currently documents deleted inline commands such as `SubmitMint` and omits the live `/admin/freeze-schedules` feature. This makes the specification contradict `MintCommand`, the durable job chain, and the admin API.

## Checks

- `git diff --check`
- verified every current `MintCommand` variant is documented and obsolete variants are absent
- verified the scheduled-freeze endpoint is present in both source and spec
- verified `AGENTS.md` remains below 40,000 bytes

## Summary by CodeRabbit

* **Documentation**
  * Updated Mint Flow guidance to reflect the latest transaction preparation, persistence, broadcasting, confirmation, and callback tracking steps.
  * Replaced outdated command-based terminology with the current job and event workflow.

PR #324, opened 2026-08-10
refactor: remove retired Fireblocks integration

findolor:
## Motivation

The retired Fireblocks integration still left runtime, CLI, dependency, deployment, and secret-management surface in the repository. RAI-1676 removes that surface while preserving permanent historical event replay and the vendor-neutral receipt-custody engine.

Closes [RAI-1676](https://linear.app/makeitrain/issue/RAI-1676/remove-the-retired-fireblocks-integration).

Closes [RAI-1711](https://linear.app/makeitrain/issue/RAI-1711/remove-fireblocks)

## Solution

- Replace the retired migration guidance with the neutral custody specification, deployment-hold runbook, and Fireblocks decommission record; rename the four custody scenarios to neutral outgoing/incoming-custodian terminology.
- Remove the temporary custody CLI drivers, Fireblocks implementation, rollback preflight, migration-only transaction plumbing, SDK dependency, Nix hashes, encrypted key files, and deployment wiring.
- Preserve custody commands, events, state, reconciliation/quiescence, `migrate_vault_receipts`, `OnchainReceiptCustody`, `CorroboratedRecipient`, `confirm_custody_holder`, and legacy `FireblocksSubmitted` / `fireblocks_tx_id` replay compatibility.
- Rebuild required event-sourced projections before custody work; bind migration authorization to the canonical/provider chain, tokenized-asset listing, execution database, corroborated recipient, and recorded custody route; reject self-destination, unrelated origins, unconfirmed custody, and batches above the proven 14-receipt bound.
- Guard deployment service-control applications while a deployment hold is armed, remove stale host plaintext key material during activation, and intentionally leave the neutral engine without a CLI driver until RAI-1681.

At public Base block 49,821,087, the retired source `0x1c66D6708914C40239D54919320b4C48cAE3D1A9` held zero of the four pinned legacy ERC-1155 receipt groups and zero of the stranded tCOIN vault/share balance. The exact contracts, identifiers, routing history, and evidence are retained in `docs/runbooks/fireblocks-decommission.md`; no pre-removal asset sweep remains.

Operator follow-ups are limited to removing Fireblocks variables from the encrypted environment payloads, deploying and verifying both profiles, revoking the external API user, and removing the obsolete TAP rule and whitelists.

## Checks

By submitting this for review, I am confirming I have done the following:

- [x] added comprehensive test coverage for the changed custody invariants and production startup path
- [x] made this PR as small as possible while preserving the six requested logical commits
- [x] linked the relevant Linear issue
- [x] ran `nix flake check`
- [x] created a disposable database and ran all migrations in the Nix shell
- [x] ran `cargo test --workspace`
- [x] ran `cargo fmt --all -- --check`
- [x] ran `cargo clippy --workspace --all-targets --all-features -- -D clippy::all -D warnings`
- [x] evaluated production and staging deploy profiles and NixOS toplevel derivations
- [x] ran targeted custody and historical replay tests
- [x] completed final `FIREBLOCKS_`, retired CLI, secret-path, and compatibility-reference sweeps
- [x] completed simplification and three staged local review rounds, supplemented with `opencode-go/deepseek-v4-flash`

Local x86_64-linux NixOS toplevel builds are unavailable from the aarch64-darwin host; remote CI provides the required target build coverage.

## Summary by CodeRabbit

- **New Features**
  - Added the `force-complete-redemption` issuer command.
  - Updated `burn-excess` with internal and external modes using service-configured network access.
  - Receipt custody migration now validates identities, ownership, balances, and routes, with safe idempotent reruns.

- **Removed Features**
  - Removed vendor-specific custody migration, verification, authentication, and related issuer commands and configuration.

- **Documentation**
  - Added deployment-hold and retired-custody cleanup runbooks.
  - Clarified migration safeguards, reconciliation, recovery, and operational procedures.


findolor:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/324?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#324** <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/324?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/324?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`
<hr />

You must have a Graphite account in order to use the merge queue. Sign up using [this link](https://app.graphite.com/invite/github/ST0x-Technology?ref=merge-queue-instructions-in-stack-comment&prId=324).

<h2></h2>
<sub>

</sub>

<sub>

</sub>

<h2></h2>

docs/runbooks/deploy-hold.md
```diff
@@ -0,0 +1,45 @@
...
+## Release the hold
+
+Release the hold only after the maintenance procedure and post-operation checks
+are complete:
+
+```sh
+rm /run/st0x/st0x-issuance.hold
+touch /run/st0x/st0x-issuance.ready
+systemctl start st0x-issuance
+```
```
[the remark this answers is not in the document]
findolor: Fixed on the latest head. Releasing the hold now removes the marker and reruns the deployment activation, which owns validation, marker creation, and service startup. The runbook explicitly forbids manually creating the readiness marker or starting the unit and verifies both afterward.

docs/runbooks/deploy-hold.md
```diff
@@ -0,0 +1,45 @@
...
+
+Verify that the unit is active and startup reconciliation completed without
+errors before resuming dependent services.
```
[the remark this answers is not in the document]
findolor: Fixed on the latest head. The runbook now covers validation and activation failures: leave failed validation inactive, rely on deploy-rs rollback for a failed profile, re-arm the hold and stop the unit if release fails, inspect status and journal output, and escalate without bypassing the marker.

src/receipt_inventory/migration.rs
```diff
@@ -870,7 +1095,21 @@ pub async fn confirm_custody_holder<P: Provider + Clone + Send + Sync>(
...
+
+async fn confirm_custody_holder_for_identity<
+    P: Provider + Clone + Send + Sync,
+>(
+    pool: &Pool<Sqlite>,
+    provider: P,
+    identity: VaultIdentity<'_>,
+    holder: Address,
+) -> anyhow::Result<usize> {
+    let VaultIdentity { network: _, chain_id, vault, underlying } = identity;
```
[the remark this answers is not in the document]
findolor: Fixed on the latest head. The custody bootstrap tests now run Anvil with the canonical Base chain ID and construct VaultIdentity through the same network, provider, and tokenized-asset listing corroboration used by the retained engine, so the fixtures no longer represent an impossible identity.

findolor:
Addressed the review-body finding outside the changed lines as well: the synthetic tokenized-asset CLI failure now says `burn transaction polling timed out`. Remaining Fireblocks-named fixtures are limited to intentional historical event/serde replay compatibility.

JuaniRios:
@coderabbitai full review

docs/runbooks/deploy-hold.md
```diff
@@ -0,0 +1,84 @@
...
+the service.
+
+## Arm the hold
+
+Create the runtime directory and arm the hold before stopping the service. This
+ensures a concurrent deploy observes the hold and leaves the service stopped:
+
+```sh
+install -d /run/st0x
+touch /run/st0x/st0x-issuance.hold
```
0xgleb, line 15: should fix: arming the hold creates `/run/st0x/st0x-issuance.hold` and stops the unit, but it never removes `/run/st0x/st0x-issuance.ready`, and that marker is the unit's only systemd-level start gate (`ConditionPathExists = markerFile` in `nix/upgradeable-services.nix`). For the whole window between arming and the next deploy, `systemctl start st0x-issuance` therefore succeeds, and `nix run .#prodServiceStart`, `.#prodServiceRestart` and `.#prodDbReset` all re-touch the marker and start the unit without consulting the hold at all. Please add `rm -f /run/st0x/st0x-issuance.ready` to the arm and re-arm blocks, add `test ! -f /run/st0x/st0x-issuance.ready` to the confirmation, and name those three flake apps here as commands that must not be run while the hold is armed. Without that, the release check `test -f /run/st0x/st0x-issuance.ready` also cannot distinguish a fresh activation from a leftover marker.

src/tokenized_asset/cli.rs
```diff
@@ -136,27 +80,6 @@ enum IssuerCommand {
     /// other redemption may already claim it. Completes the redemption and
     /// settles its receipt reservation like a normal burn confirmation.
     ForceCompleteRedemption(Box<ForceCompleteRedemptionArgs>),
```
0xgleb, line 90: should fix: this removes `sweep-legacy-receipts` together with the only in-repo record of what it was built to move: `LEGACY_HOLDER` 0x1c66D670, `TURNKEY_RECIPIENT`, `LIQUIDITY_BOT`, `TCOIN_VAULT`, the four pinned ERC-1155 receipt groups that predate inventory tracking, and `STRANDED_TCOIN` (0.2 tCOIN). The retained engine cannot substitute, because `quiescent_tracked_holdings` builds its batch only from `inventory.receipts_with_balance()`, so untracked receipts are structurally invisible to it, and no function in `migration.rs` moves ERC-20 shares at all. The same change deletes the encrypted Fireblocks keys and the SDK while the PR body still lists sweeping leftover wallet assets as an open operator follow-up, so please confirm on-chain that those four receipt balances and the tCOIN balance at the legacy holder are zero before merging, and note it in the description. If any is non-zero, run the sweep on the pre-removal revision first; if they are all zero, move the addresses, receipt ids and the non-obvious routing rule (stranded tCOIN goes to the liquidity bot rather than the redemption wallet, because redemption detection skips senders it cannot attribute) into `docs/runbooks/` so the follow-up stays executable after the API user is revoked.

tests/receipt_custody.rs
```diff
@@ -296,21 +351,29 @@ async fn run_operator_migration(
...
+    let identity = VaultIdentity::verify(
+        &pool,
+        provider,
+        Network::Base,
+        chain_id,
+        vault,
+        &underlying,
+    )
+    .await?;
+    confirm_custody_holder(&pool, provider, identity, outgoing).await?;
```
0xgleb: minor: this call, and the matching one in the rollback test, is load-bearing, and `confirm_custody_holder` has zero production callers after this change. In production the only writer of custody is `ReceiptReconciler::confirm_custody`, reachable solely from `run_startup_reconciliation`, which returns early on an empty inventory, and these tests start the service before any receipt exists, so no real pass ever confirms custody; without this line `migrate_vault_receipts` would fail `corroborate_recorded_custody_route` with `CustodyUnobserved`. The suite therefore proves the engine works once a test-only entry point recorded custody, not that a real deployment ever arrives at migration with custody recorded, which is exactly the seam this PR creates. Restarting the outgoing service after the receipts are discovered and asserting custody is already `Held { holder: outgoing_wallet }` would cover it.

src/receipt_inventory/migration.rs
```diff
@@ -707,89 +766,257 @@ pub async fn migrate_vault_receipts<P: Provider + Clone + Send + Sync>(
...
-    .await?;
+    recipient: Address,
+) -> Result<RecordedCustodyRoute, MigrationRefusal> {
+    match recorded {
+        None => Err(MigrationRefusal::CustodyUnobserved { vault }),
+        Some(recorded) if recorded == holder => {
+            Ok(RecordedCustodyRoute::AtHolder)
+        }
+        Some(recorded) if recorded == recipient => {
+            Ok(RecordedCustodyRoute::AtRecipient)
```
0xgleb, line 991: minor: this arm accepts the request whenever the recorded holder equals the requested recipient, and nothing corroborates the requested `holder` on that branch. `CorroboratedRecipient::verify` only proves the recipient is non-zero, differs from the holder, and exists on-chain, and the inventory's `moved_from()` is never consulted, so passing an arbitrary source X together with the real destination R resolves to `AtRecipient`, reads zero at X, classifies everything as migrated, and reports `MigrationOutcome::AlreadyMigrated { receipts: N }` for a route that never existed. Nothing is corrupted only because `RecordCustodyMigration` dedupes on the destination alone, which is accidental rather than designed. Consider threading the recorded origin in and refusing this arm unless `recorded_origin == Some(holder)`, with a unit test alongside `a_recorded_destination_allows_an_idempotent_rerun`.

SPEC.md
```diff
@@ -1191,16 +1177,15 @@ backfiller has already checkpointed past the deposits that created them.
...
-backfiller, startup and periodic reconciliation, and any future caller — goes
-through this one handler, so no code path can apply a wrong-wallet reading. The
-refusal is per vault and writes nothing: the service keeps serving vaults whose
-custody matches while a displaced vault fails loudly at ERROR. This is what
-makes a single-asset cutover safe — vaults that have not migrated yet are
-refused rather than wiped — and it is also what lets the service operate
-normally after a bulk cutover while a straggler vault awaits its own migration.
-A pass that cannot read every balance confirms nothing, so one flaky call cannot
+destructive zero reading outright while no holder has ever been confirmed.
+Custody is confirmed automatically once a reconciliation pass reads every
```
0xgleb: minor: read in place, three lines above the unchanged mention of startup and periodic reconciliation, this says the running service arms the displacement guard, and it does not. `ConfirmCustody` is only ever issued by `ReceiptReconciler::confirm_custody` inside `reconcile()`, which runs once at boot from `run_startup_reconciliation`; the periodic loop runs `ReceiptBackfiller` and never confirms a holder. That startup call is also logged and continued in `lib.rs`, and a tracked receipt reading zero while custody is `Unobserved` aborts the pass with `CustodyUnconfirmed` before `confirm_custody` is reached, so custody can stay unobserved with no retry until the next restart. Since this sentence is replacing the command that was the documented remedy, please name startup reconciliation specifically and state that nothing retries it.

src/receipt_inventory/mod.rs
```diff
@@ -944,8 +932,8 @@ pub(crate) enum ReceiptInventoryError {
     #[error(
         "Refusing balance reading for receipt {receipt_id} taken against \
          wallet {observed_wallet}: recorded custody holder is {holder}. \
-         Point the service at the holder or migrate custody with \
-         `issuer migrate-receipts`."
+         Point the service at the holder or use the approved custody-migration \
```
0xgleb: minor: this refusal and `CustodyUnconfirmed`, plus `ReconcileError::CustodyDisplaced`, now point operators at an approved custody-migration procedure while the same change deletes every implementation of one (`migrate-receipts`, `confirm-custody`, `verify-custodians`) along with `docs/turnkey-receipt-migration-plan.md`, and the new runbook covers only deploy holds. The guard is also self-perpetuating: `apply_reading` propagates the aggregate error out of `reconcile()` before `confirm_custody()` runs, so the pass that would arm custody is the pass being refused, and `lib.rs` only logs the startup error and keeps serving. With no driver until RAI-1681, an on-call engineer following this text has no next step, and `AGENTS.md` in this same diff states the engine deliberately has no CLI driver, which directly contradicts it. Either keep a minimal entry point over the surviving `confirm_custody_holder` and `migrate_vault_receipts`, or make these three messages name the remedy that exists today and say plainly that no custody move is possible yet.

src/receipt_inventory/migration.rs
```diff
@@ -664,38 +716,45 @@ impl CorroboratedRecipient {
...
+        write!(formatter, "{}", self.recipient)
     }
 }
 
 /// Migrates one vault's receipt custody to `recipient`, end to end.
 ///
-/// Public so the operator CLI and the cutover end-to-end test drive the same
-/// code: a test exercising a hand-rolled transfer would prove nothing about
-/// what the operator actually runs.
+/// Public so a future operator driver and the receipt-custody end-to-end test
```
0xgleb, line 773: minor: `migrate_vault_receipts`, `confirm_custody_holder`, `recorded_migration_origin`, `VaultIdentity`, `CorroboratedRecipient` and `MigrationOutcome` stay `pub` while every operator entry point is deleted, so the only consumer left is `tests/receipt_custody.rs` and `dead_code` cannot see the situation. That freezes a semver-relevant surface for a driver that does not exist yet, and this PR widens it (`VaultIdentity::verify`, a holder-carrying `CorroboratedRecipient`, `corroborate_recorded_custody_route`), against the repo's own rule to keep visibility as restrictive as possible. Worth noting the deleted driver also derived the holder and the signer from one value so they could never disagree, and nothing in the engine preserves that pairing now. Either land the driver alongside the engine, or demote to `pub(crate)` plus whatever the integration test genuinely needs and move the scenario coverage into `src/`.

src/receipt_inventory/migration.rs
```diff
@@ -870,7 +1095,21 @@ pub async fn confirm_custody_holder<P: Provider + Clone + Send + Sync>(
     identity: VaultIdentity<'_>,
     holder: Address,
 ) -> anyhow::Result<usize> {
-    let VaultIdentity { chain_id, vault, underlying } = identity;
+    identity.corroborate_provider(&provider).await?;
+    identity.corroborate_listing(pool).await?;
```
0xgleb, line 1150: minor: the public `confirm_custody_holder` now runs `corroborate_provider` and `corroborate_listing` before any work, but the `custody_bootstrap` tests were switched to `confirm_custody_holder_for_identity`, which skips the wrapper entirely. `pool_with_migrations` and `seeded_vault` seed no tokenized-asset listing, so those tests would hit `ListingNotFound` at the public boundary real operators have to pass, meaning a regression in the listing lookup, the pool wiring, or the error mapping could ship green. One test calling the public entry point against a pool built like `vault_identity_corroboration::pool_with_listing`, plus a wrong-listing case asserting the refusal, would cover the wiring.

src/receipt_inventory/migration.rs
```diff
@@ -707,89 +766,257 @@ pub async fn migrate_vault_receipts<P: Provider + Clone + Send + Sync>(
...
+impl<'a> VaultIdentity<'a> {
+    /// Corroborates a complete vault identity against the provider and the
+    /// tokenized-asset listing.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error when the provider reports another chain, the listing
+    /// does not exist, the listing names another vault, or its projection
+    /// cannot be loaded.
+    pub async fn verify<P: Provider>(
```
0xgleb, line 835: minor: `VaultIdentity::verify` is the DB-backed gate every real caller goes through, yet no test calls it and no fixture seeds a pool without a listing, so the `ListingNotFound` refusal is never exercised; `vault_identity_corroboration` only drives `from_observations` and `corroborate_listing` against pools that always hold some listing. Separately, `chain_id` is fully derivable from `network` through the total `const fn Network::chain_id()`, so a network paired with a chain it does not name is representable in a brand-new public signature and rejected at runtime instead of being unrepresentable; deriving it inside the constructor would let `corroborate_network_chain`, `NetworkChainMismatch` and its test go away. Also, `corroborate_network_chain` runs twice per call, at line 811 and again in `from_observations`, and the pre-I/O rationale the deleted CLI carried in a comment did not come along, so a one-line note would stop someone collapsing the wrong one.

deploy.nix
```diff
@@ -83,25 +83,9 @@ let
...
-          else
-            rm -f "$fireblocks_key_tmp"
-            exit 1
-          fi
-        ''
-
         # Validate config + secrets before restarting. If validation fails,
         # deploy-rs exits non-zero and rolls back instead of starting bad config.
-        "set -a; . ${cfg.decryptedEnvPath}; set +a; DATABASE_URL=sqlite:///mnt/data/issuance.db ENVIRONMENT=${deploymentEnvironment} FIREBLOCKS_SECRET_PATH=${cfg.decryptedFireblocksKeyPath} ${cfg.profilePath}/bin/validate-config"
+        "set -a; . ${cfg.decryptedEnvPath}; set +a; DATABASE_URL=sqlite:///mnt/data/issuance.db ENVIRONMENT=${deploymentEnvironment} ${cfg.profilePath}/bin/validate-config"
```
0xgleb, line 92: minor: the activation step that decrypted `fireblocks-secret-issuance-${env}.key.age` into `/run/agenix/fireblocks-secret-issuance.key` is gone, but nothing removes the plaintext key earlier deploys already wrote. `/run/agenix` is populated exclusively by these explicit `rage` and `install` commands, with no `age.secrets` module, tmpfiles rule or `ExecStartPre` managing it, so on a long-lived prod or staging host that 0440 `root:st0x` RSA key stays readable by the group the internet-facing unit runs as until a reboot. Add an idempotent `rm -f /run/agenix/fireblocks-secret-issuance.key` to this activation list for one release cycle, or record removing that exact path in a surviving runbook next to revoking the API user. With the `.age` sources also deleted, nothing else in the repo will remind an operator the file exists.

docs/runbooks/deploy-hold.md
```diff
@@ -0,0 +1,84 @@
...
+```sh
+rm /run/st0x/st0x-issuance.hold
+```
+
+Then, from the deployment checkout, run the service deployment again so profile
+activation validates the configuration, creates the readiness marker, and starts
+the unit:
+
+```sh
+nix run .#prodDeployService -- st0x-issuance
```
0xgleb: minor: this is the step that ends the maintenance window, and it is the only operator-facing deploy invocation in the repo that omits the SSH identity; `docs/nixos-provisioning.md` passes `-i "$SSH_IDENTITY"` everywhere, and only CI relies on the default path because its setup writes exactly `~/.ssh/id_ed25519`. `deployPreamble` reaches `parseIdentity`, which falls back to `$HOME/.ssh/id_ed25519` and exits if it is absent, and that identity is needed both to SSH to the droplet and to decrypt `infra/.remote-prod.age` for host resolution, so an operator whose infra key lives elsewhere fails right here with the hold already removed and the unit still stopped. Document it as `nix run .#prodDeployService -- -i "$SSH_IDENTITY" st0x-issuance` and note that the flag must precede the profile argument, because the preamble consumes it before the prelude reads `$1`.

src/receipt_inventory/migration.rs
```diff
@@ -707,89 +766,257 @@ pub async fn migrate_vault_receipts<P: Provider + Clone + Send + Sync>(
...
+            provider.get_chain_id().await.map_err(ReceiptCustodyError::from)?;
+
+        corroborate_provider_chain(self.chain_id, provider_chain_id)
+    }
+
+    async fn corroborate_listing(
+        self,
+        pool: &Pool<Sqlite>,
+    ) -> anyhow::Result<()> {
+        let listed_vault = find_vault(pool, self.underlying, &self.network)
```
0xgleb, line 896: minor: `corroborate_listing` and `VaultIdentity::verify` resolve the listing through `find_vault`, a raw `tokenized_asset_view` read, and the caller that used to guarantee the projection was healed is deleted. `AssetAdmin::connect` ran `prepare_event_sourced_startup::<TokenizedAsset>` first precisely because this code runs while the service is deliberately stopped, and its comment records production carrying that view empty for weeks. On a database whose view was recreated by a migration and not yet healed, this now reports `ListingNotFound` for an asset the event log holds, and the quiescence gates fail the other way, since empty mint and redemption views report zero in-flight work and the gate passes. Either run the same startup hygiene inside the engine, or take an already-prepared store handle instead of a raw `Pool` so a caller cannot supply an unhealed database, plus a test that clears the view tables.

docs/runbooks/deploy-hold.md
```diff
@@ -0,0 +1,84 @@
+# Deployment hold
+
+Use the deployment hold when maintenance requires the issuance service to stay
```
0xgleb, line 3: minor: `docs/fireblocks-migration-authorization.md`, deleted here, was the only in-repo record of the external authority this integration required, including a TAP rule with no function-selector field that permits the vault account to call any method on the whitelisted Receipt contract, and its own instruction to remove that rule and the whitelist entry once every vault had migrated. Removing the consumer does not remove the grant, so after this change nothing in the repository will prompt anyone to revoke it, and a future reader has no way to learn which console objects and credentials this system depended on. A short decommission note in this directory or in `secret/`, listing the whitelisted contract entry, the TAP rule, the API user, and the host-side key path with the current state of each, would keep that recoverable; a link to an external tracker from the touched files would do as well.

src/receipt_inventory/migration.rs
```diff
@@ -1729,9 +1798,9 @@ impl<P: Provider + Clone + Send + Sync> ReceiptCustody
         self.check_vault(permit.vault())?;
         ensure_permit_covers(permit, holdings)?;
 
-        // A previous rollback may have been broadcast just before this CLI
-        // crashed. While its transaction remains pending, re-submitting would
-        // create a second attempt without knowing the first one's outcome.
+        // A previous transfer may have been broadcast just before the driver
```
0xgleb, line 1897: minor: deleting `FireblocksReceiptCustody` also deletes `MAX_RECEIPTS_PER_ENGINE_TRANSFER` and the chunked submission loop, leaving this `transfer_custody` as the only path, building one `safeBatchTransferFrom` over every tracked identifier with no cap, no size check and no gas derivation, since `MigratableHoldings` carries an unbounded `Vec` and `quiescent_tracked_holdings` only filters zero balances. The deleted doc held the only recorded evidence at this boundary, that 240 receipts is the observed failure and batches up to 14 are proven to pass, and nothing replaces it while the new comment in `reconcile_holdings` asserts one atomic batch positively; the real chain has only ever been exercised with a single receipt. The per-chunk `ensure_holder_quiescent` and certification and owner-freeze re-reads went with the loop too, so a large move now has one permission snapshot for the whole transfer. Either cite the bound (Base's block gas limit and a measured per-identifier cost) in this doc comment, or bound the batch and keep `reconcile_holdings`' per-identifier classification as the documented resume path, so an unmovable vault is a typed `MigrationRefusal` rather than an estimation failure discovered mid-window.

src/receipt_inventory/migration.rs
```diff
@@ -1,37 +1,33 @@
...
 //! ([`super::reconcile`], reached from `run_startup_reconciliation`). Once
 //! custody has moved but the service is still configured with the outgoing
-//! signer, every one of those reads returns zero, which reconciles as a
-//! depletion and drops the receipts from the aggregate outright. Freezing the
-//! underlying does not prevent this: a freeze rejects new mints, it does not
-//! stop the service or serialize against it. The order is stop the service,
-//! migrate, swap the signer configuration, start again.
+//! signer, every one of those reads returns zero. The retained custody guard
+//! refuses that pass without writing a depletion, but the deployment must
+//! still be quiescent so mint and redemption workers cannot race the move.
```
0xgleb, line 18: minor: this paragraph still requires the issuer service to be stopped and the deployment quiescent, but the mandatory confirmation in `run_forward_via_fireblocks` and `run_rollback_transfer` was the only thing that ever stood behind it, and its own comment called that the one precondition nothing downstream can detect. What remains reads persisted state only: `require_quiescent` counts reserved burns and stuck mints and redemptions, and `ensure_holder_quiescent` compares nonce views, so a mint or redemption arriving after the store read, or a deposit landing a fresh receipt at the outgoing wallet after the batch is built, passes both gates. The new runbook defines hold and readiness markers that the engine never consults. Requiring an explicit witness in `migrate_vault_receipts`, either marker state or a store-level maintenance lease the running service holds, would put the precondition somewhere a caller cannot skip.

src/receipt_inventory/migration.rs
```diff
@@ -605,36 +646,41 @@ pub enum MigrationOutcome {
...
     ///
     /// # Errors
     ///
-    /// Returns an error for the zero address, and when the chain has no record
-    /// of the address at all.
+    /// Returns an error for the zero address, the current holder, and when the
+    /// chain has no record of the address at all.
     pub async fn verify<P: Provider>(
         provider: &P,
+        holder: Address,
```
0xgleb, line 713: minor: after this change `CorroboratedRecipient::verify` is the only destination gate and it passes on a non-zero nonce or any native balance, while `rollback_gas_reserve` and `require_rollback_gas` (the 0.001 ETH minimum whose comment warned that a bare non-zero check lets one wei through and leaves a real rollback unbroadcastable) and `verify_rollback_signing` (the sign-only proof of control) are both deleted with nothing re-homed into the engine. Was that deliberate on the grounds that RAI-1681 targets a contract destination, where a sign-only rollback proof has no analogue and a code plus ERC-1155-receiver check replaces it? If so, is anything tracking an address-appropriate proof of control for the future signer-rotation driver this module doc still anticipates, given that a dust-funded but uncontrolled destination is exactly the case the deleted reserve was sized for?

src/receipt_inventory/migration.rs
```diff
@@ -1516,8 +1586,7 @@ async fn verify_custody_moved(
 }
 
 /// Reads both nonce views for `holder` and rejects a non-quiescent wallet —
-/// the single entry point for the read+check pair, so every submission path
-/// (custody transfers, chunked Fireblocks calls, the legacy-receipt sweeps)
+/// the single entry point for the read+check pair, so every custody transfer
```
0xgleb, line 1685: nit: `ensure_holder_quiescent`, and likewise `require_quiescent`, `reconcile_holdings` and `migrate_vault_custody`, are all still `pub(crate)` but now have no reference anywhere outside this module, since the sweep CLI and the Fireblocks custody impl that reached them are both gone. Dropping the qualifier lets the compiler tell you when one stops being used, which is the reason the repo asks for the narrowest visibility, and the `#[cfg(test)]` module is a child so it keeps access either way. Worth re-checking the `MigratableHoldings` and `TransferPermit` accessors on the same pass.

flake.nix
```diff
@@ -170,16 +170,14 @@
...
 
-        # Vendoring reads the whole workspace lock, so the main crate's two git
-        # deps (event-sorcery/sqlite-es + fireblocks-sdk) need pinned hashes
-        # even though the dto crate never pulls them in.
+        # Vendoring reads the whole workspace lock, so the main crate's git
+        # dependency needs a pinned hash even though the dto crate never pulls
+        # it in.
         cargoVendorDir = craneLib.vendorCargoDeps {
           src = crateSrc;
           outputHashes = {
```
0xgleb, line 186: nit: this attrset is spelled out byte-identically here and in `rust.nix:27-30` with no shared binding, and this PR had to edit both in lockstep to drop the `fireblocks-sdk-rs` entry, which is the failure mode the duplication creates. A git dep added to only one of the two makes the other `vendorCargoDeps` call fall back to `builtins.fetchGit`, which resolves only `refs/heads/main` and fails the build on any tag or feature-branch rev, a long way from its cause; divergent hashes are silent in the same way. `rust.nix` is already `callPackage`'d, so passing the map in from `flake.nix`, or importing one `nix/cargo-git-hashes.nix` in both, would make them agree by construction.

src/vault/mod.rs
```diff
@@ -47,9 +47,9 @@ pub(crate) trait VaultService: Send + Sync {
     /// The returned bytes and hash must be persisted before calling
     /// [`VaultService::submit_mint`].
     ///
-    /// `external_tx_id` is a signing-backend correlation id (Fireblocks
-    /// idempotency / Turnkey labeling). It is **not** vault-direct double-mint
-    /// protection — after the first `MintTxIntended`, only exact-hash
+    /// `external_tx_id` is a legacy signing-backend correlation id. It is
```
0xgleb: nit: the signing-backend correlation role is genuinely gone, but the value is not vestigial: it round-trips into `MintEvent::TxSubmitted`, and `Mint::next_retry_attempt` reads it back through `latest_known_external_tx_id` and parses the `-retry-N` suffix off it to derive the next retry's identifier. Calling it legacy is precisely what invites a later cleanup to stop threading it, which silently resets retry-attempt derivation and makes `retry_submission_external_tx_id` produce the same id for every attempt. Better to state the remaining role, that no signing backend correlates on it any more and its only live consumer is `Mint::next_retry_attempt`, keeping the existing note that it is not double-mint protection.

src/tokenized_asset/cli.rs
```diff
@@ -1571,93 +324,6 @@ async fn verified_chain_id(
     Ok(chain_id)
```
0xgleb, line 821: nit: `verified_chain_id`'s doc comment still claims it is reached only once the operator has confirmed intent, so an `--rpc-url` typo cannot surface before they have decided. That was true of the custody drivers this change deletes, but the sole remaining caller, `run_force_complete_redemption`, resolves the chain before it ever reaches its confirmation prompt, the opposite ordering. Replace the ordering sentence with the rationale that is still true, that the RPC is the authority on which chain is reached while only the operator knows which chain they meant, and a deterministic deployment can place the same vault address on two chains.

src/receipt_inventory/migration.rs
```diff
@@ -664,38 +716,45 @@ impl CorroboratedRecipient {
...
 ///
-/// Public so the operator CLI and the cutover end-to-end test drive the same
-/// code: a test exercising a hand-rolled transfer would prove nothing about
-/// what the operator actually runs.
+/// Public so a future operator driver and the receipt-custody end-to-end test
+/// can drive the same reusable engine.
 ///
-/// `provider` must be signing as `holder`, the wallet whose custody moves —
-/// ERC-1155 only lets the holder or an approved operator move a balance.
+/// `provider` must be signing as the witness's holder, the wallet whose custody
```
0xgleb: nit: `the witness's holder` is not defined in this doc comment, the module doc, or the `CorroboratedRecipient` type doc, so a reader of this comment alone cannot resolve which value they must sign as. That matters because this is the safety-critical instruction about who must control the signing key, aimed at exactly the future driver author the doc anticipates. Something self-contained like `recipient`'s corroborated holder, the wallet whose custody moves, would read without prior knowledge of the type.

src/receipt_inventory/migration.rs
```diff
@@ -707,89 +766,257 @@ pub async fn migrate_vault_receipts<P: Provider + Clone + Send + Sync>(
...
+}
+
+const fn corroborate_recorded_route_observation(
+    route: RecordedCustodyRoute,
+    observed: &SourceCustody,
+    vault: Address,
+    holder: Address,
+    recipient: Address,
+) -> Result<(), MigrationRefusal> {
+    if matches!(route, RecordedCustodyRoute::AtRecipient)
```
0xgleb: nit: this reduces the two-variant `RecordedCustodyRoute` to a boolean, and `AtHolder` is constructed but never matched anywhere, so a third variant would be treated as permitted here with no compiler error; the same applies to `matches!(observed, SourceCustody::Holds(_))`. Since this is the last check before an irreversible batch transfer is authorised, the failure direction on an unhandled variant is permissive. An exhaustive `match (route, observed)` would make adding a variant a compile error rather than an implicitly allowed migration.

.env.secrets.example
```diff
@@ -90,20 +90,6 @@ TURNKEY_API_PRIVATE_KEY=<hex-encoded-p256-api-private-key>
...
-# NOT set FIREBLOCKS_SECRET_PATH here — the deploy activation injects the
-# decrypted RSA key's path, and a stale copy in this file would shadow it.
-FIREBLOCKS_API_USER_ID=<fireblocks-api-user-uuid>
-
-# Defaults ("0", "8453:BASECHAIN_ETH", "production") — set only if they differ:
-# FIREBLOCKS_VAULT_ACCOUNT_ID=0
-# FIREBLOCKS_CHAIN_ASSET_IDS=8453:BASECHAIN_ETH
-# FIREBLOCKS_ENVIRONMENT=production
-
 ISSUER_API_KEY=${ISSUER_API_KEY}
```
0xgleb, line 93: nit: this file is the documented source for `secret/st0x-issuance-{env}.env.age` per `docs/nixos-provisioning.md`, but the template is trimmed here without re-encrypting either env file, so whatever `FIREBLOCKS_*` entries they carry keep being decrypted to `/run/agenix/st0x-issuance.env` and exported through `EnvironmentFile`, the `issuer` wrapper, and the deploy-time `validate-config` call, now with no reader anywhere in the tree. The template and the deployed secret therefore describe different variable sets, and the identifier needed to locate the API user still due for revocation stays in the service environment and in any environment dump. Re-encrypting both files without the `FIREBLOCKS_*` entries as part of this change, or committing it as an explicit decommission step, would close the drift.

src/receipt_inventory/migration.rs
```diff
@@ -643,6 +689,12 @@ impl CorroboratedRecipient {
             );
         }
 
+        if recipient == holder {
```
0xgleb, line 722: nit: `execute_migration` takes the whole migration source from `recipient.holder()`, and nothing ever compares it to the address the `Provider` actually signs as, so a future driver that corroborates the right holder but attaches the wrong signing provider fails inside `transfer_custody` as an opaque contract error rather than a typed refusal. During a custody move that is the wrong kind of failure to diagnose under time pressure, and it invites repeated submissions while nonces are in flux. Comparing `holder` against the provider's signer when the provider is wallet-backed, and refusing with a named `MigrationRefusal`, would keep it on the typed path; the approved-operator case can be documented.

src/receipt_inventory/migration.rs
```diff
@@ -605,36 +646,41 @@ pub enum MigrationOutcome {
     },
 }
 
-/// A destination the chain itself has already seen.
+/// A custody route whose destination the chain itself has already seen.
```
0xgleb: nit: the doc still says existence implies corroboration, but the type is now a three-field custody route and only `recipient` is actually checked; `holder` gets the `recipient != holder` comparison and nothing else at construction. The safety consequence looks bounded, because `corroborate_recorded_custody_route`, `reconcile_holdings` and `ensure_permit_covers` all re-validate `holder` on the single path to a transfer, `corroborate_recipient_chain` reconciles the stored `chain_id` against the migration's own, and `holder()` and `chain_id()` are private to this module. Still worth scoping the invariant sentence to the destination, so a reader does not infer a guarantee about the other two fields.

SPEC.md
```diff
@@ -926,33 +926,19 @@ against the local SQLite store, and is where future issuer actions (e.g. `mint`,
 - `issuer freeze <UNDERLYING>` — dispatch the `Freeze` command.
 - `issuer unfreeze <UNDERLYING>` — dispatch the `Unfreeze` command.
 - `issuer status <UNDERLYING>` — print the underlying's current freeze status.
+- `issuer force-complete-redemption <ISSUER_REQUEST_ID> --burn-tx-hash
```
0xgleb: minor: `ForceCompleteRedemptionArgs` also requires `--network`, `--chain-id` and `--reason`, none of which have defaults, so this documented invocation fails clap validation as written. It also drops the safety property the code documents on `chain_id`, that the command refuses unless `--network`, `--chain-id` and the RPC all name the same chain so a destructive terminalization needs two independent statements of where it runs, and the paragraph that recorded that scoping is deleted in this same hunk. Either extend the bullet to the full required form and restore a sentence about the cross-check and where `--rpc-url` comes from, or drop it, since documenting this command is unrelated to removing the retired integration.

SPEC.md
```diff
@@ -926,33 +926,19 @@ against the local SQLite store, and is where future issuer actions (e.g. `mint`,
...
+  <TX_HASH>`
```
0xgleb: nit: this bullet documents a command that already exists at this commit and is untouched by the diff, so it is a change not required by the task. Small, but it costs the PR the property of reading as nothing but a removal, which is most of its review value. Worth splitting into its own documentation change.

0xgleb (changes requested):

docs/runbooks/deploy-hold.md
findolor: Fixed. Arming and re-arming now remove the readiness marker before stopping the unit, verification requires the hold to exist, readiness to be absent, and the unit to be inactive, and the start/restart/db-reset flake apps refuse while the hold is armed.

src/tokenized_asset/cli.rs
findolor: Fixed. I verified the retired source balances at public Base block 49,821,087; the four pinned ERC-1155 receipt groups and stranded tCOIN balance were all zero. The exact source, contracts, identifiers, routing history, evidence, and remaining revocation steps are now preserved in docs/runbooks/fireblocks-decommission.md and summarized in the PR description.

tests/receipt_custody.rs
findolor: Fixed. The scenarios no longer call confirm_custody_holder directly. They restart the outgoing service after receipt discovery so production startup reconciliation records custody, and migration success now proves that production path.

src/receipt_inventory/migration.rs
findolor: Fixed. The idempotent destination path now requires the aggregate recorded origin to equal the caller-stated source and returns a typed CustodyOriginMismatch otherwise, with a regression test.

SPEC.md
findolor: Fixed. The specification now states that custody confirmation occurs during startup reconciliation only, is not retried periodically, and requires correcting the cause and restarting if startup reconciliation fails.

src/receipt_inventory/mod.rs
findolor: Fixed. The custody errors now state plainly that no operator CLI exists until RAI-1681 and direct operators to keep the service stopped and escalate instead of referring to a nonexistent procedure.

src/receipt_inventory/migration.rs
findolor: Retaining this public boundary is intentional under the approved PR contract: migrate_vault_receipts, confirm_custody_holder, VaultIdentity, CorroboratedRecipient, and MigrationOutcome are the reusable vendor-neutral engine that must survive removal of the temporary CLI. The engine remains tested externally while no CLI driver is exposed until RAI-1681. Leaving this thread open because I did not make the requested visibility change.

src/receipt_inventory/migration.rs
findolor: Fixed. Tests now call the public VaultIdentity::verify and confirm_custody_holder boundaries using a real listing-backed pool, including listing failure coverage, rather than only exercising internal helpers.

src/receipt_inventory/migration.rs
findolor: Fixed the boundary coverage: VaultIdentity::verify is now exercised for a missing listing and for projection rebuild. The explicit chain id remains intentional as an independent operator/provider cross-check; the pre-I/O duplicate check is documented so it is not accidentally collapsed.

deploy.nix
findolor: Fixed. Deployment activation now idempotently removes /run/agenix/fireblocks-secret-issuance.key, eliminating plaintext left by earlier deployments.

docs/runbooks/deploy-hold.md
findolor: Fixed. The release step now uses nix run .#prodDeployService -- -i "$SSH_IDENTITY" st0x-issuance and explains that the identity flag must precede the profile argument.

src/receipt_inventory/migration.rs
findolor: Fixed. Custody entry points now rebuild the tokenized-asset, mint, redemption, and receipt-inventory projections from the event store before safety reads. Regression tests clear those projections and prove listing and in-flight work are reconstructed.

docs/runbooks/deploy-hold.md
findolor: Fixed. Added docs/runbooks/fireblocks-decommission.md covering the API user, TAP rule, whitelists, host key path, known legacy asset routing, verified balances, and remaining operator actions.

src/receipt_inventory/migration.rs
findolor: Fixed. The neutral engine now refuses more than 14 nonzero receipts with a typed ReceiptBatchTooLarge error before submission, documents the historical observed bound, and retains reconciliation as the resume path. A regression test proves no transfer is attempted.

src/receipt_inventory/migration.rs
findolor: The generic engine cannot portably inspect a host deployment marker, and this PR intentionally exposes no production driver. Deployment service-control apps now machine-enforce the hold, and SPEC makes hold verification a mandatory RAI-1681 driver acceptance criterion. Leaving this open for RAI-1681 because the engine signature itself was not changed.

src/receipt_inventory/migration.rs
findolor: This is deliberately destination-backend-specific and belongs to the absent RAI-1681 driver. SPEC now requires an address-appropriate destination proof and gas readiness before that driver may invoke the engine; the generic engine continues to prove on-chain existence and post-transfer balances. Leaving this open for RAI-1681.

src/receipt_inventory/migration.rs
findolor: Fixed. Internal migration helpers, holdings, permit accessors, and traits were narrowed to private visibility; only the approved reusable public engine boundary remains public.

flake.nix
findolor: I am keeping this out of scope. The contract required the Fireblocks dependency and both matching Nix hashes to be removed atomically; introducing a shared Nix hash abstraction would be an unrelated flake refactor. Leaving this thread open for a dedicated cleanup.

src/vault/mod.rs
findolor: Fixed. The documentation now describes external_tx_id as the live retry-attempt identifier consumed by Mint::next_retry_attempt, while clarifying that no signing backend correlates on it and it is not duplicate-mint protection.

src/tokenized_asset/cli.rs
findolor: Fixed. The comment now states the surviving rationale: the RPC reports the reached chain, the operator states the intended chain, and deterministic deployments can share an address across chains; it no longer claims confirmation occurs first.

src/receipt_inventory/migration.rs
findolor: Fixed. The instruction now names the corroborated outgoing holder explicitly as the wallet whose custody moves.

src/receipt_inventory/migration.rs
findolor: Fixed. Authorization now uses an exhaustive match over RecordedCustodyRoute and SourceCustody, so future variants fail compilation instead of being permissively accepted.

.env.secrets.example
findolor: Fixed through the explicit decommission path requested in the comment. The new runbook gives exact commands for editing both encrypted environment payloads without exposing them, removing all FIREBLOCKS variables, deploying, verifying the live service environment, and revoking the API user.

src/receipt_inventory/migration.rs
findolor: The generic engine intentionally supports both holder signing and approved ERC-1155 operators, so it cannot require the provider address to equal the holder. SPEC now requires the future RAI-1681 driver to bind its concrete signer or approved operator before invocation. Leaving this open for that driver.

src/receipt_inventory/migration.rs
findolor: Fixed. The invariant documentation is now scoped to the corroborated destination only and no longer implies that existence alone validates the source holder or chain.

SPEC.md
findolor: Fixed by removing the unrelated force-complete-redemption documentation bullet from this PR.

SPEC.md
findolor: Fixed by removing the unrelated command documentation so the specification diff remains scoped to Fireblocks retirement.

docs/runbooks/fireblocks-decommission.md
```diff
@@ -0,0 +1,52 @@
...
+   remaining automation depends on them.
+4. Remove the temporary Transaction Authorization Policy rule and its operator
+   group membership. Then remove each Receipt contract's Base asset entry and
+   delete the contract-wallet whitelist container if it has no other use.
+5. Review the retired wallet's complete asset inventory and transaction history
+   for anything outside the explicit zero-balance table above. Escalate any
+   unexplained balance or destination before closing the workspace account.
+
+Policy cleanup must follow asset verification: deleting authorization first can
+strand an item that was missed by the known legacy table.
```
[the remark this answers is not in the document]
findolor: Fixed. The full retired-wallet inventory and transaction-history review is now step 3, while the API user and policy still exist; revocation and policy cleanup follow only after that review.

infra/default.nix
```diff
@@ -284,10 +284,15 @@ let
...
+          if ! ssh_remote "test ! -e /run/st0x/st0x-issuance.hold"; then
+            echo "ERROR: deployment hold is armed; refusing to reset the database" >&2
+            exit 1
+          fi
+
           _restart_service() {
             if [ "$stay_stopped" = false ]; then
               echo "Ensuring st0x-issuance is restarted on ${env}..." >&2
-              ssh_remote "mkdir -p /run/st0x && touch /run/st0x/st0x-issuance.ready && systemctl start st0x-issuance" || true
+              ssh_remote "test ! -e /run/st0x/st0x-issuance.hold && mkdir -p /run/st0x && touch /run/st0x/st0x-issuance.ready && systemctl start st0x-issuance" || true
```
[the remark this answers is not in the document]
findolor: Fixed. The EXIT restart path now emits a clear warning and deliberately leaves the service stopped when a hold appears during reset. The normal restart path uses the explicit fatal guard and distinguishes an armed hold from SSH or systemctl failure.

SPEC.md
```diff
@@ -929,30 +929,12 @@ against the local SQLite store, and is where future issuer actions (e.g. `mint`,
...
-`--network` / `--chain-id` (cross-checked against the RPC-reported chain); its
-RPC endpoint is **not** a CLI flag — it uses the service environment for that
-network (`CHAIN_<NETWORK>_RPC_URL`, with legacy `RPC_URL` as Base fallback), the
-same secrets the long-running bot loads.
+
+`burn-excess` is listing/network-scoped and takes `--network` / `--chain-id`
+(cross-checked against the RPC-reported chain). Its RPC endpoint is **not** a
+CLI flag — it uses the service environment for that network
+(`CHAIN_<NETWORK>_RPC_URL`, with legacy `RPC_URL` as Base fallback), the same
+configuration the long-running bot loads.
```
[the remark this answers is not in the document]
findolor: Fixed. The duplicate burn-excess paragraph was removed and its unique CHAIN_<NETWORK>_RPC_URL and legacy Base RPC_URL details were moved into the shared-arguments paragraph.

SPEC.md
```diff
@@ -1117,27 +1098,36 @@ addresses as free-form operator targets):**
...
-itself, and must be corroborated by the chain (an address with no transaction
-history and no native balance is what a corrupted config value looks like) via a
-witness type the transfer cannot be reached without.
+### Receipt custody
+
+The receipt-moving engine is vendor-neutral and deliberately has no operator
+entry point. It remains reusable library code until
+[RAI-1681](https://linear.app/makeitrain/issue/RAI-1681) supplies a dedicated
+driver with a signing provider and stated destination. The engine and custody
+state are exercised by end-to-end tests in `tests/receipt_custody.rs`.
```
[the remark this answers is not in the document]
findolor: Fixed. Added a sibling Freeze, unfreeze, and status heading so those CLI semantics are no longer part of the Receipt custody section.

SPEC.md
```diff
@@ -1191,17 +1179,16 @@ backfiller has already checkpointed past the deposits that created them.
...
+destructive zero reading outright while no holder has ever been confirmed.
+Custody is confirmed automatically when startup reconciliation reads every
+tracked balance without error; `confirm_custody_holder` remains available to a
+future custody driver for holders that are not the signing wallet. Confirmation
+is not retried periodically: after a startup read failure, correct the cause and
+restart the service. Every balance reader goes through this handler, so no code
+path can apply a wrong-wallet reading. The refusal is per vault and writes
+nothing: the service keeps serving vaults whose custody matches while a
+displaced vault fails loudly at ERROR. A pass that cannot read every balance
+confirms nothing, so one flaky call cannot disarm the guard.
```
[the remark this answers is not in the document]
findolor: Fixed. The specification now states that an empty inventory confirms no custody, leaves the vault Unobserved, and that confirmation requires at least one successfully read tracked receipt.

tests/receipt_custody.rs
```diff
@@ -189,49 +241,53 @@ async fn mint_before_cutover(
...
     )
     .await?;
     harness::wait_for_mock_hits(mint_callback_mock, 2).await?;
 
-    Ok(U256::from(25) * U256::from(10).pow(U256::from(18)))
+    Ok(CanaryMint {
+        issuer_request_id,
+        shares: U256::from(25) * U256::from(10).pow(U256::from(18)),
+    })
 }
```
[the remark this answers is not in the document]
findolor: Fixed. CANARY_QUANTITY is now the single source for both the submitted decimal quantity and expected 18-decimal shares.

tests/receipt_custody.rs
```diff
@@ -296,21 +351,29 @@ async fn run_operator_migration(
...
+        chain_id,
+        vault,
+        &underlying,
+    )
+    .await?;
+    confirm_custody_holder(&pool, provider, identity, outgoing).await?;
     let outcome =
-        migrate_vault_receipts(&pool, provider, identity, outgoing, incoming)
-            .await?;
+        migrate_vault_receipts(&pool, provider, identity, incoming).await?;
```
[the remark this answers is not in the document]
findolor: Fixed. Both direct confirm_custody_holder calls and the import were removed. Migration success now depends on production startup reconciliation having recorded the outgoing holder; all four scenarios pass with that path.

tests/receipt_custody.rs
```diff
@@ -566,47 +637,47 @@ async fn test_wallet_cutover_redeems_pre_cutover_receipt_after_restart()
...
-        turnkey_withdrawals, 1,
+        incoming_withdrawals, 1,
         "the new wallet must submit exactly one historical-receipt burn"
     );
     assert_eq!(
-        fireblocks_provider.get_transaction_count(turnkey_wallet).await?,
+        outgoing_provider.get_transaction_count(incoming_wallet).await?,
         2,
         "the new wallet must sign exactly one canary mint and one historical-receipt burn"
     );
```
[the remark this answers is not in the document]
findolor: Fixed. Removed the unused post-migration deposit so the remaining withdrawal and nonce assertions stay focused on the historical-receipt burn.

0xgleb (approved):

JuaniRios (approved):

infra/default.nix
```diff
@@ -221,7 +221,7 @@ let
           ${requireDecryptedSecrets}
           echo "Starting st0x-issuance on ${env}..."
           ssh ''${identity:+-i "$identity"} "root@$host_ip" \
-            "mkdir -p /run/st0x && touch /run/st0x/st0x-issuance.ready && systemctl start st0x-issuance"
+            "test ! -e /run/st0x/st0x-issuance.hold || { echo 'ERROR: deployment hold is armed; refusing to start st0x-issuance' >&2; exit 1; }; mkdir -p /run/st0x && touch /run/st0x/st0x-issuance.ready && systemctl start st0x-issuance"
```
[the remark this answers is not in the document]
rouzwelt: deffered to a followup issues/PR

PR #333, opened 2026-08-13
[RAI-1740] Drop subgraph dependency fomr issuance bot

ueco-jb:
The issuance bot's only use of the subgraph is [OaSchemaCache](https://github.com/ST0x-Technology/st0x.issuance/blob/a1369aeabb9a00e92ddfee2525a76a0282a2d641/src/vault/rain_meta.rs#L276-L284).
It queries Goldky for the vault's receiptVaultInformation and extracts the OA schema hash (IPFS CID) embedded in mint receipt metadata.

Replace the subgraph query with a direct read: one eth_getLogson the vault address for the ReceiptVaultInformation event and take the latest.
No new failure modes: on failure the mint proceeds with metadata omitting OA_SCHEMA, same as today. Fixed mode (admin excess burn) is unchanged.
Remove entries from the config, clap args, subgraph schema validation, env examples, runbook row and the SPEC mention.

Acceptance:
- OaSchemaCache resolves the hash from the on chain event with no subgraph_url configured
- schema hash decode roundtrip covered by a mocked provider test
- config tests updated: no subgraph var in any chain group
- Eth runtime boots with no subgraph configured

---
<a href="https://app.blacksmith.sh/ST0x-Technology/codesmith/st0x.issuance/pr/333"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789226814&installation_model_id=19370&pr_number=333&repository=ST0x-Technology%2Fst0x.issuance&return_to=https%3A%2F%2Fgithub.com%2FST0x-Technology%2Fst0x.issuance%2Fpull%2F333&signature=0923370f5bf199bb1eb246fa0b91ff22d18339cde3783525453d1c0c80e39bcb"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


ueco-jb:
I will leave it in draft for now, waiting for RAI-1215 to be merged first.

ueco-jb:
https://app.graphite.com/github/pr/ST0x-Technology/st0x.issuance/333/changes#comment-PRRC_kwDOQDN0ts7iswDw  
I added a 20 second deadline: on timeout it omits OA_SCHEMA and retries next call. The scan result is now cached even when no event is found, so the history walk runs once per vault per process instead of on every mint. This is safe because that None means a confirmed absence (the scan reached block 0 and found nothing).  
I did not floor at backfill_start_block: it is set near activation, but the `ReceiptVaultInformation` event is emitted earlier at deployment; flooring there would bring back the silent miss.   
Scanning to block 0 and caching the empty result keeps the cost bounded, the full scan runs at most once per vault, so I avoid hardcoding each vault's deployment block just to set a floor.  
  
I added a regression test: after a cached "no schema" result (a valid one, the schema really not being present), a second lookup returns None without rescanning.

ueco-jb:
All rabits were put to sleep. (the outside diff comments included) ^

JuaniRios (changes requested):
Looks good to drop the Goldsky dependency. I think we should change the cache behavior though:

1. Persist the latest OA schema per vault and the last scanned block.

2. On startup, resume from that checkpoint. If none exists, backfill once.

3. Keep polling ReceiptVaultInformation events forward as new blocks arrive.

This catches schema changes while the bot is running and avoids walking backwards from the chain tip during a mint.

receiptVaultInformation is permissionless, so newest-event-wins is an attack surface. I’m okay handling that in a follow-up PR, but we should track it and only accept events from trusted senders.

I also checked rain-metadata. It now includes the OA magic numbers, but it still rejects the additional OA_SCHEMA CBOR key. So it can’t replace our custom codec yet.

Also checking if OA SCHEMA is necessary for the issuance bot, but that's another discussion outside this PR

ueco-jb:
> Also checking if OA SCHEMA is necessary for the issuance bot, but that's another discussion outside this PR

You mean the liquidity bot? I already checked it, there are 0 references for both subgraph and: oa_schema, OaSchema, OA_HASH_LIST, OA_STRUCTURE, IPFS CID, cbor/ciborium, rain_meta, or receiptVaultInformation  
(I think those are all possible variants I noticed)

ueco-jb:
> Looks good to drop the Goldsky dependency. I think we should change the cache behavior though:
>
> 1. Persist the latest OA schema per vault and the last scanned block.
> 2. On startup, resume from that checkpoint. If none exists, backfill once.
> 3. Keep polling ReceiptVaultInformation events forward as new blocks arrive.
>
> This catches schema changes while the bot is running and avoids walking backwards from the chain tip during a mint.
>
> receiptVaultInformation is permissionless, so newest-event-wins is an attack surface. I’m okay handling that in a follow-up PR, but we should track it and only accept events from trusted senders.
>
> I also checked rain-metadata. It now includes the OA magic numbers, but it still rejects the additional OA_SCHEMA CBOR key. So it can’t replace our custom codec yet.

​Agreed on dropping Goldsky.   
I will create a follow up PR stacked on top of this one though, because this PR only needs to drop the subgraph dependency, and the checkpoint model is a bigger change that builds on top of the read it adds.  
\- persist the latest schema per vault  
\- resume from a last scanned block  
\- poll ReceiptVaultInformation forward

The trusted sender filter goes in the same follow up, since `receiptVaultInformation` is permissionless and the walk currently takes the newest event from any emitter.

rain metadata: agreed, we keep the custom codec since the crate still rejects the extra OA_SCHEMA key. I'll add a comment at the parser.

ueco-jb:
Cancelled in favor of https://github.com/ST0x-Technology/st0x.issuance/pull/357

